PR #33, opened 2020-03-17
Add ability to set up some simple logging via Session

vinceatbluelabs:
I needed some logging while figuring out https://github.com/bluelabsio/records-mover/pull/34.  So I added some!


records_mover/log_levels.py
```diff
@@ -24,7 +23,3 @@ def set_levels(app_name: str,
     # Log all messages to stderr
     root_logger = logging.getLogger()
     root_logger.setLevel(adjusted_log_level(default_root_log_level, app_name))
-
-    # Log app messages to job_log by default
-    app_logger = logging.getLogger(app_name)
-    app_logger.setLevel(adjusted_log_level(default_app_log_level, app_name))
```
vinceatbluelabs: This was a leftover from BlueHQ job logging - we could set the log level of things that went to job_context.logger separately, and that was the logger which would actually go to the web console..

records_mover/records/cli.py
```diff
@@ -101,6 +99,7 @@ def main() -> None:
     if func is None:
         parser.print_help()
     else:
-        log_levels.set_levels(name)
-        cli_logging.basic_config()
+        from records_mover import Session
+        session = Session()
+        session.set_stream_logging()
```
vinceatbluelabs: Creating a throwaway session object here is kind of awkward.  I suppose we could create set_stream_logging() as an independent function you can call outside of a session (in addition to being a method on the session?)/

records_mover/session.py
```diff
@@ -156,6 +158,19 @@ def _boto3_session(self) -> boto3.session.Session:
         else:
             return self.creds.boto3_session(self._default_aws_creds_name)
 
+    def set_stream_logging(self,
+                           app_name: str = 'records_mover',
+                           default_root_log_level: int = logging.INFO,
+                           stream: IO[str] = sys.stdout,
+                           fmt: str = '%(asctime)s - %(message)s',
+                           datefmt: str = '%H:%M:%S') -> None:
```
vinceatbluelabs: Lots of arguments, but all are defaulted, so you can just call it and get some kind of sensible logging set up if you want.

.circleci/config.yml
```diff
@@ -101,7 +101,7 @@ jobs:
       - store_artifacts:
           path: test-reports
       - store_artifacts:
-          path: cover.tar.gz
+          path: cover # TODO .tar.gz
```
bluelabsbutler: Uncompleted punchlist item detected--consider resolving or moving this to your issue tracker

records_mover/session.py
```diff
@@ -156,6 +158,34 @@ def _boto3_session(self) -> boto3.session.Session:
...
+        :param fmt: Logging format to send to Python'slogging.Formatter() - determines what details
+         will be sent.
+        :param datefmt: Date format to send to Python'slogging.Formatter() - determines how the
+        current date/time will be recorded in the log.
+        """
+        log_levels.set_levels(app_name, default_root_log_level)
+        formatter = logging.Formatter(fmt, datefmt)
+        handler = logging.StreamHandler(stream=stream)
+        handler.setFormatter(formatter)
+        root_logger = logging.getLogger()
```
cwegrzyn: It looks like this is setting application-wide logging, rather than logging for just the records_mover package. In contrast, [the boto logging helpers](https://github.com/boto/boto3/blob/develop/boto3/__init__.py#L37) add a logging handler only to the named logger (which defaults to the boto3 logger). I would suggest we follow that pattern. If folks desire app-wide logging, they can just call logging.basicConfig-- seems like this helper should be targeted towards helping people log just records mover output easily. What do you think?

records_mover/records/cli.py
cwegrzyn: Fwiw, boto3 seems to have the logging helper outside of the session object, in contrast to botocore which has it in the session object (but doesn't actually make use of `self` in any meaningful way and so could just as well live as a top-level function like in boto3)

records_mover/__init__.py
```diff
@@ -1,4 +1,13 @@
-from .types import JsonSchema  # noqa
```
vinceatbluelabs: I'm not sure why this was here, but the tests and mypy are fine with me removing it, so probably no better time!

records_mover/records/cli.py
vinceatbluelabs: I'm gonna propose we do both here.

records_mover/session.py
vinceatbluelabs: Makes sense!

PR #31, opened 2020-03-14
Initial MySQL support

vinceatbluelabs:
This adds support for MySQL!

Other non-obvious changes inside:
* Document new driver installation process!  I'm probably missing some things, and this is most certainly not to the level of a "step-by-step", but it's a start!
* Adds some testing support for databases that, like MySQL have unsigned types.  It seemed to fit in pretty smoothly to the existing numeric support!
* The usual battle against mypy coverage.  I got a pretty big gain by adding a typestub for logging, and found some cases where we were handing an exception object as the first argument to e.g. logger.info().  When tested, that outputs an empty line, not any exception information.
* With CSV sources that turn into dataframes, we hadn't really been handling the date/time columns as anything other than strings.  That's been good enough up until this point because our target databases have been pretty generous with the parsing (generally even with INSERTs you can just throw it in a string in a number of common formats).  However, MySQL requires a real date/time object on insert to a date/time column.  I changed our CSV import code to use the records schema to identify and parse the date/time columns, and changed our pandas->CSV export code to use the right format in that direction as well.

Note that this isn't yet passing bigfiles, but I believe is passing everything else.  I want to take a crack at addressing that (mostly from integration test growth, I think), but wanted to have a cleaner diff in the meantime.  Maybe I'll do a PR against this branch once this branch is OKed to keep a clean diff.

PR #35, opened 2020-04-03
Clean up table_validator.py

vinceatbluelabs:
I was working through some integration test failures in https://github.com/bluelabsio/records-mover/pull/31 and found the current table validator class in the integrations tests here needed some love to make it more explicit and clear.

Let me know if it looks like a positive move!  It's blowing up bigfiles because table_validator.py is longer.  I might consider extracting out the timezone processing prediction logic into its own class, but thought it'd make sense to get an initial review before I took that on.  I might end up doing that as part of #31 depending on the timing.


tests/integration/records/table_validator.py
```diff
@@ -29,13 +39,13 @@ def database_default_store_timezone_is_us_eastern(self):
...
 
-    def supports_time_without_date(self):
+    def supports_time_without_date(self) -> bool:
         # for shame, redshift!
         return (self.engine.name != 'redshift'
                 and (self.source_data_db_engine is None or
                      self.source_data_db_engine.name != 'redshift'))
 
-    def variant_doesnt_support_seconds(self, variant):
+    def variant_doesnt_support_seconds(self, variant: DelimitedVariant):
```
bluelabsbutler, line 48: F821 undefined name 'DelimitedVariant'

tests/integration/records/table_validator.py
```diff
@@ -1,20 +1,44 @@
...
+from sqlalchemy.sql.elements import TextClause
+from typing import Optional, Dict, Any, Union
 from .timezone import set_session_tz
 from .expected_column_types import expected_column_types
+from records_mover.records import DelimitedVariant
 
 
 logger = logging.getLogger(__name__)
 
+# TODO: Can I pass in load_variant to constructor?
```
bluelabsbutler, line 15: Uncompleted punchlist item detected--consider resolving or moving this to your issue tracker

Makefile
```diff
@@ -11,6 +11,7 @@ clean: coverageclean typecoverageclean
 
 typecheck:
 	mypy --cobertura-xml-report typecover --html-report typecover .
+	mypy tests/integration/records
```
vinceatbluelabs, line 14: I wanted to use types to document complex parts of the integration tests like this, and of course validate those annotations, but not have it count against the metrics.

records_mover/records/types.py
```diff
@@ -58,6 +58,8 @@ class UrlDetailsEntry(TypedDict):
                                           total=False)
 
     RecordsFormatType = Literal['delimited', 'parquet']
+
+    DelimitedVariant = Literal['dumb', 'csv', 'bigquery', 'bluelabs', 'vertica']
```
vinceatbluelabs, line 62: There's room to expand use of this into records mover proper, but it'll require a bit of code moving to do the validation up front - maybe a follow-up PR.

tests/integration/records/multi_db/test_records_table2table.py
```diff
@@ -96,18 +95,15 @@ def move_and_verify(self, source_dbname, target_dbname, variant_used_internally)
...
 
 if __name__ == '__main__':
+    set_stream_logging()
+
     for source in DB_TYPES:
         for target in DB_TYPES:
-            variant_used_internally = 'bluelabs'
-            if 'vertica' in [source, target]:
-                variant_used_internally = 'vertica'
-
```
vinceatbluelabs, line 110: You'll notice this never gets passed in 😆 

It's now the responsibility of `table_validator.py` to figure out which variant is exported from the source database and imported to the target database.  It's the responsibility of `test_records_table2table.py` to tell which databases are being moved between.

tests/integration/records/single_db/test_records_load.py
```diff
@@ -146,5 +146,5 @@ def load(self, format_type, variant, hints={}, broken=False, sourcefn=None):
             self.assertIn(out.move_count, [None, 1])
 
     def verify_db_table(self, variant):
-        validator = RecordsTableValidator(self.engine)
-        validator.validate(variant, schema_name=self.schema_name, table_name=self.table_name)
+        validator = RecordsTableValidator(self.engine, file_variant=variant)
```
vinceatbluelabs, line 149: I'm clarifying with the terminology 'file_variant' that this is a source file (which may be loaded into pandas and turned into some other variant for loading - the `load_variant`).

Note that as Parquet support gets more fully integrated and tested we'll want to change these variant arguments into records_format arguments.

tests/integration/records/table_validator.py
```diff
@@ -155,9 +203,28 @@ def validate_data_values(self, variant, schema_name, table_name):
...
            not self.database_default_store_timezone_is_us_eastern()):
-            # Example date that we'd be loading:
+            #
+            # In this case, we correctly tell Pandas that we have are
+            # at noon:34 US/Eastern, and tell Pandas to format the
+            # datetime format.  Unfortunately, if you don't specify a
+            # timezone as part of that format, Pandas just prints the
+            # TZ-naive hour.
+            #
+            utc_hour = 12
```
vinceatbluelabs, line 216: This case was was sort of handled by coincidence before, and I think was confusing things quite a bit when debugging test failures.

vinceatbluelabs:
(this was merged into the mysql branch)

PR #37, opened 2020-04-10
Work around sqlalchemy interaction with sqlalchemy-redshift

vinceatbluelabs:
See https://github.com/sqlalchemy-redshift/sqlalchemy-redshift/issues/195


vinceatbluelabs:
See https://circleci.com/workflow-run/1820a629-d9a3-48b2-a395-821e94700ac3 for an example of the problem

PR #39, opened 2020-04-14
Strengthen directory validator validation

vinceatbluelabs:


tests/integration/records/directory_validator.py
```diff
@@ -48,25 +52,33 @@ def validate_records_schema(self):
...
+                acceptable_field_types_by_db = {
+                    # Redshift doesn't support TIME type:
+                    # https://docs.aws.amazon.com/redshift/latest/dg/r_Datetime_types.html
+                    'redshift': [
+                        'integer', 'string', 'string', 'string',
+                        'string', 'string', 'string', 'date', 'string',
+                        'datetime', 'datetimetz'
+                    ]
+                }
+                if actual_field_types == acceptable_field_types_by_db[self.source_db_type]:
```
cwegrzyn, line 74: Not a big deal, this will raise an exception if, say, there is a regression in vertica field handling. Obviously that yields the same result as raising an assertion, but with a different message.

Something like this would let it make it to the assertion:

```suggestion
                if actual_field_types == acceptable_field_types_by_db.get(self.source_db_type):
```

vinceatbluelabs: Sounds good!

PR #51, opened 2020-05-04
Fix potential issue in PostgreSQL CSV format handling

vinceatbluelabs:
It looks like a bunch of tests pretended to validate that all hints passed to the code were handled - but they weren't actually validating that!

This fixes those tests, as well as fixes one bug in the Postgres driver revealed by that.

### DIFF
--- a/records_mover/db/postgres/copy_options/csv.py
+++ b/records_mover/db/postgres/copy_options/csv.py
@@ -95,6 +95,7 @@ def postgres_copy_options_csv(unhandled_hints: Set[str],
             cant_handle_hint(fail_if_cant_handle_hint, 'quoting', hints)
     else:
         _assert_never(mode)
+    quiet_remove(unhandled_hints, 'quoting')
 
     # As of the 9.2 release (documentation as of 2019-03-12), there's
     # no statement in the docs on what newline formats are accepted in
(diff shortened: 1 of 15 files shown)


records_mover/db/postgres/copy_options/csv.py
```diff
@@ -95,6 +95,7 @@ def postgres_copy_options_csv(unhandled_hints: Set[str],
             cant_handle_hint(fail_if_cant_handle_hint, 'quoting', hints)
     else:
         _assert_never(mode)
+    quiet_remove(unhandled_hints, 'quoting')
```
vinceatbluelabs, line 98: It looks like we weren't clearing this out in this path, which one of the now-fixed unit tests revealed.  Not sure why this wasn't seen before!

records_mover/db/postgres/unloader.py
```diff
@@ -104,7 +104,6 @@ def known_supported_records_formats_for_unload(self) -> List[BaseRecordsFormat]:
     def can_unload_this_format(self, target_records_format: BaseRecordsFormat) -> bool:
         try:
             unload_plan = RecordsUnloadPlan(records_format=target_records_format)
-            unhandled_hints = set()
```
vinceatbluelabs, line 107: This was a copy and pasto - I initialize it shortly after this.

PR #58, opened 2020-05-08
Add cli-extra-test

vinceatbluelabs:
It turns out a dependency on numpy (which isn't installed in the default extras) had snuck in.  Here's an example of things failing:

```sh
$ pyenv uninstall -f mvrec && pyenv virtualenv 3.8.2 mvrec && pyenv shell mvrec
[...]
$ pip install records-mover[cli]
[...]
$ mvrec
Traceback (most recent call last):
  File "/Users/broz/.pyenv/versions/mvrec/bin/mvrec", line 6, in <module>
    from records_mover.records.cli import main
  File "/Users/broz/.pyenv/versions/3.8.2/envs/mvrec/lib/python3.8/site-packages/records_mover/__init__.py", line 14, in <module>
    from . import records
  File "/Users/broz/.pyenv/versions/3.8.2/envs/mvrec/lib/python3.8/site-packages/records_mover/records/__init__.py", line 18, in <module>
    from .schema import RecordsSchema
  File "/Users/broz/.pyenv/versions/3.8.2/envs/mvrec/lib/python3.8/site-packages/records_mover/records/schema/__init__.py", line 2, in <module>
    from .schema import RecordsSchema  # noqa
  File "/Users/broz/.pyenv/versions/3.8.2/envs/mvrec/lib/python3.8/site-packages/records_mover/records/schema/schema/__init__.py", line 4, in <module>
    from ..field import RecordsSchemaField
  File "/Users/broz/.pyenv/versions/3.8.2/envs/mvrec/lib/python3.8/site-packages/records_mover/records/schema/field/__init__.py", line 2, in <module>
    import numpy as np
ModuleNotFoundError: No module named 'numpy'
(mvrec) broz@graybookpro:~$
```

This PR reproduces this problem using an integration test and fixes it.

Doing this involved exploring some of the tension between extras and requirements.txt.  I'd be curious of your take on the balance to strike here.


.circleci/config.yml
```diff
@@ -542,12 +557,17 @@ workflows:
           filters:
             tags:
               only: /v\d+\.\d+\.\d+(-[\w]+)?/
+      - cli-extra-test:
+          name: cli-extra-test
+#          requires:  # TODO
```
bluelabsbutler, line 562: Uncompleted punchlist item detected--consider resolving or moving this to your issue tracker

setup.py
```diff
@@ -107,10 +107,14 @@ def initialize_options(self) -> None:
...
 
-itest_dependencies = (
-    [
-        'jsonschema',  # needed for directory_validator.py
-    ] +
+nose_dependencies = [
+    'nose'
+]
+
+itest_dependencies =  [
```
bluelabsbutler, line 114: E222 multiple spaces after operator

setup.py
```diff
@@ -222,7 +226,22 @@ def initialize_options(self) -> None:
     'google-cloud-storage'
 ] + smart_open_dependencies
 
-unittest_dependencies = (
+
+typecheck_dependencies = [
+    'mypy==0.770',
+    'lxml', # needed by mypy HTML coverage reporting
```
bluelabsbutler, line 232: E261 at least two spaces before inline comment

.circleci/config.yml
```diff
@@ -43,7 +50,7 @@ commands:
               fi
             fi
       - save_cache:
-          key: deps-v1-<<parameters.python_version>>-<<parameters.pandas_version>>-<<parameters.extras>>-{{ .Branch }}-{{ checksum "requirements.txt" }}-{{ checksum "setup.py" }}
+          key: deps-v1-<<parameters.python_version>>-<<parameters.pandas_version>>-<<parameters.extras>>-<<parameters.include_dev_dependencies>>-{{ .Branch }}-{{ checksum "requirements.txt" }}-{{ checksum "setup.py" }}
```
vinceatbluelabs, line 53: `requirements.txt` includes numpy-stubs, which transitively depends on numpy.  So let's only include it when we're actually type checking...

.circleci/config.yml
```diff
@@ -315,6 +325,24 @@ jobs:
...
+          pandas_version: ''
+          # we want this just like a user would install it, not with
+          # dev tools installed
+          include_dev_dependencies: false
+      - run:
+          name: Run tests
+          command: |
+            . venv/bin/activate
+            mvrec --help
+            mvrec
```
vinceatbluelabs, line 345: Just a simple test to make sure basic stuff works with *just* the CLI extra.

.circleci/config.yml
```diff
@@ -356,7 +384,7 @@ workflows:
               only: /v\d+\.\d+\.\d+(-[\w]+)?/
       - test:
           name: test-3.8
-          extras: '[unittest]'
+          extras: '[unittest,typecheck]'
```
vinceatbluelabs, line 387: I took this as an opportunity to further pull things out of requirements.txt into appropriate extras - including mypy under the new typecheck extra.

records_mover/records/schema/field/constraints/constraints.py
```diff
@@ -115,7 +116,7 @@ def from_sqlalchemy_type(required: bool,
         return RecordsSchemaFieldConstraints(required=required, unique=unique)
 
     @staticmethod
-    def from_numpy_dtype(dtype: np.dtype,
+    def from_numpy_dtype(dtype: 'np.dtype',
                          unique: bool) -> 'RecordsSchemaFieldConstraints':
         from .string import RecordsSchemaFieldStringConstraints
         from .integer import RecordsSchemaFieldIntegerConstraints
```
vinceatbluelabs, line 122: These two files were bugs reproduced by the new integration test.

requirements.txt
```diff
@@ -3,17 +3,17 @@ setuptools>34.3.0
...
 # this is pre-release, but seems to work!  Let's keep it pinned to a
 # commit until they make a release, just in case they break
 # compatibility.  Feel free to try to update!
+#
+# Note: it looks like numpy 1.20 is likely to roll these in natively
+# without a need for a separate stub package:
+#
+#  https://github.com/numpy/numpy/blob/master/doc/release/upcoming_changes/16558.new_feature.rst
+#  https://github.com/numpy/numpy/pull/16558
 git+https://github.com/numpy/numpy-stubs.git@be9a479#egg=numpy-stubs
```
vinceatbluelabs, line 19: Unfortunately, I wasn't able to pull this one in to an extra, as there's no published package.  But it sounds like the next numpy will come with stubs by default!

requirements.txt
```diff
@@ -3,17 +3,17 @@ setuptools>34.3.0
...
 twine
 flake8
-nose
 nose-progressive
-coverage
-mock
 pytz
-mypy==0.770
-lxml # needed by mypy HTML coverage reporting
 pylint
```
vinceatbluelabs, line 8: The rest of the stuff here is:

1. things we need on the command-line but never in CI (flake8, nose-progressive)
2. stuff we need at package publishing time - maybe we could add another extra fo rthose?  (wheel, twine)
3. cruft that's obsolete or assuredly covered by something else given integration tests pass and these don't get included transitively (pytz, pylint)
4. numpy-stubs (see note below)

requirements.txt
```diff
@@ -3,17 +3,15 @@ setuptools>34.3.0
...
 nose-progressive
-coverage
-mock
-pytz
-mypy==0.770
-lxml # needed by mypy HTML coverage reporting
-pylint
-# sqlalchemy-vertica-python==0.4.5
-sqlalchemy-stubs>=0.3
+
```
vinceatbluelabs, line 7: I needed to pull nose out to an extra so we could include it during integration testing without including everything here.  At that point, it also made sense to pull out coverage and mock and mypy to match the pattern.  pylint and pytz appear to be old cruft that's no longer required.

The rest of the stuff here is:

1. things we need on the command-line but never in CI (flake8, nose-progressive)
2. stuff we need at package publishing time (wheel, twine)
3. numpy-stubs (see note below)

Number one and two could be dealt with via an extra - assuming we haven't taken extras too far already, maybe that's the way to go?

PR #67, opened 2020-05-26
Move default creds handling over to BaseCreds class

vinceatbluelabs:
This removes some complexity out of the Session class.


records_mover/creds/base_creds.py
```diff
@@ -1,11 +1,24 @@
 from db_facts.db_facts_types import DBFacts
-from typing import TYPE_CHECKING, Iterable
+from records_mover.database import db_facts_from_env  # TODO: Maybe move this to creds package?
```
bluelabsbutler, line 2: Uncompleted punchlist item detected--consider resolving or moving this to your issue tracker

metrics/bigfiles_high_water_mark
```diff
@@ -1 +1 @@
-983
+981
```
vinceatbluelabs, line 1: session.py is off the hotlist!

```
351: tests/integration/itest
317: records_mover/records/schema/field/__init__.py
313: setup.py
```

PR #68, opened 2020-05-27
Allow default credential objects to be passed into Session()

vinceatbluelabs:
This came from an internal request in the context of a web application, where the AWS creds for each Session would be cobbled together based on the user's configuration and passed in.

While asking the user to create a custom BaseCreds subclass and pass in an instance of it to Session() could certainly accomplish the same thing, it seems like providing the ability to pass in the various credential objects directly would be a more direct solution and address those super common simple cases where there's only one cred of a given type to wrangle.


vinceatbluelabs:
Hi @cwegrzyn - this one is relatively straightforward, and blocking a train of other changes - it'd be great if you could prioritize this one for review.  Thanks!!

PR #69, opened 2020-05-28
Allow IAM username suffix to be added to S3 scratch URL

vinceatbluelabs:
This PR retires the scratch-s3-bucket script and introduces the ability to use an IAM username as part of the scratch bucket URL with appropriate config files in place.  I'll work to get those set up so we can release this and have it work for our internal use.

I also moved the scratch bucket calculation over to BaseCreds, and renamed the functions to include the word 'default'.  Your AWS IAM username depends on which AWS creds you are using.  We always use the default AWS creds when loading/unloading on Redshift.  That might not be true in the future, and once we provide more flexibility there, we'll want to extend this code to take in the appropriate credentials name for the AWS creds.

PR #70, opened 2020-05-29
Allow configuration to be set for session type

vinceatbluelabs:
This will allow folks using custom cred types (like LastPass!) to configure this globally via config files.  Combined with #69 this will allow enterprise uses to be configured while more generic settings are used by default.


records_mover/creds/creds_via_env.py
```diff
@@ -15,6 +15,7 @@
 class CredsViaEnv(BaseCreds):
     def _gcp_creds(self, gcp_creds_name: str,
                    scopes: Iterable[str]) -> 'google.auth.credentials.Credentials':
+        # TODO: Make this the default one?
```
bluelabsbutler, line 18: Uncompleted punchlist item detected--consider resolving or moving this to your issue tracker

docs/CONFIG.md
```diff
@@ -96,7 +96,7 @@ URL or configure a TOML-style file in one of the following locations:
 
 Example file:
 
-```toml
+```ini
```
vinceatbluelabs, line 99: Turns out config_resolver uses https://docs.python.org/3/library/configparser.html which uses a simplified INI format which is not actually TOML.  TIL!

docs/CONFIG.md
```diff
@@ -109,3 +109,72 @@ downloaded for local processing) will be stored per Python's
 which allow for configuration via the `TMPDIR`, `TEMP` or `TMP` env
 variables, and generally default to
 [something reasonable per your OS](https://docs.python.org/3/library/tempfile.html#tempfile.gettempdir).
+
+## Cloud credentials (e.g., S3/GCS/Google Sheets)
```
vinceatbluelabs, line 113: Since database credentials come out of db-facts, the major difference made in switching the default creds provider from CredsViaLastPass and CredsViaEnv is how non-default GCP credentials are passed--this text should address that and help provide guidance.

records_mover/session.py
```diff
@@ -30,11 +30,20 @@ def _infer_session_type() -> str:
     if 'RECORDS_MOVER_SESSION_TYPE' in os.environ:
         return os.environ['RECORDS_MOVER_SESSION_TYPE']
 
+    config_result = get_config('records_mover', 'bluelabs')
+    cfg = config_result.config
+    if 'session' in cfg:
+        session_cfg = cfg['session']
+        session_type: Optional[str] = session_cfg.get('session_type')
```
cwegrzyn, line 37: Could `creds_type` be a more appropriate name here? Does this control anything about the session other than how it gets the creds?

PR #72, opened 2020-06-05
Add logo

vinceatbluelabs:


README.md
```diff
@@ -1,4 +1,6 @@
-# Records Mover - mvrec
+<img
+ src="https://raw.githubusercontent.com/bluelabsio/records-mover/add_logo/docs/records-mover-horizontal.png"
```
vinceatbluelabs, line 2: - [ ] TODO: Update URL to point to master.

metrics/mdl_high_water_mark
```diff
@@ -1 +1 @@
-5
+8
```
vinceatbluelabs, line 1: This doesn't look right with both a banner image and a top-level markdown header, but markdownlint both really doesn't want raw HTML and also doesn't want a missing top-level markdown header.  Oh well.

README.md
vinceatbluelabs, line 2: - [ ] TODO: Update URL to point to master.

PR #101, opened 2020-08-11
Adjust to Pandas breaking change, bump Python version

vinceatbluelabs:
Pandas 1.1 has broken Record Mover's usage of the read_csv() function by adding error checking in cases where a certain argument would be unused.

Details of the Pandas change:

* https://github.com/pandas-dev/pandas/issues/27394
* https://github.com/pandas-dev/pandas/pull/31383

See Records Mover test failures here: 
* https://app.circleci.com/pipelines/github/bluelabsio/records-mover/1089/workflows/e62f1cf0-f8d0-4e22-9652-112df72b02b8/jobs/9439


.python-version
```diff
@@ -1 +1 @@
-records-mover-3.8.3
+records-mover-3.8.5
```
vinceatbluelabs, line 1: Bump to latest Python.

PR #107, opened 2020-09-16
Handle multiple fileobjs in DoMoveFromFileobjsSource

vinceatbluelabs:
When uploading large files into BigQuery (which accepts them via a stream) we ran into two limitations:
1. `DoMoveFromFileobjsSource` doesn't know how to use `ConcatFiles` to paste together the multiple CSV files that things were turned into.
2. Google's libraries want to call `.tell()` on the stream they're handed, which is not currently supported by `ConcatFiles`  (probably the cause for #1)

This adds `tell()` support in `ConcatFiles`, allowing us to use it in `DoMoveFromFileobjsSource `


records_mover/utils/concat_files.py
```diff
@@ -18,10 +20,23 @@ def readable(self) -> bool:
         return True
 
     def readall(self) -> bytes:
-        return b''.join(f.read(-1) for f in self._files)
+        out = b''
+        while self._files:
+            f = self._files.pop(0)
+            out = out + f.read()
+        # TODO write a test that forces me to write this code
```
bluelabsbutler, line 27: Uncompleted punchlist item detected--consider resolving or moving this to your issue tracker

records_mover/records/targets/table/move_from_fileobjs_source.py
```diff
@@ -21,9 +24,9 @@ def __init__(self,
         self.fileobjs_source = fileobjs_source
         all_fileobjs = list(self.fileobjs_source.target_names_to_input_fileobjs.values())
         if len(all_fileobjs) != 1:
-            raise NotImplementedError("Teach me how to append and load more than one "
-                                      "file into same table")
-        self.fileobj = all_fileobjs[0]
+            self.fileobj = ConcatFiles(all_fileobjs)  # type: ignore
```
vinceatbluelabs, line 27: I don't know how to convince the ConcatFiles class to type as `IO[bytes]`.  It certainly is duck-typed as one.

records_mover/utils/concat_files.py
```diff
@@ -18,10 +20,23 @@ def readable(self) -> bool:
...
+    def tell(self) -> int:
+        return self._tell
 
     def read(self, size: int = -1) -> bytes:
-        if size < 0:
+        #  "When size is omitted or negative, the entire contents of
+        #  the file will be read and returned"
+        #
+        # https://docs.python.org/3/tutorial/inputoutput.html
+        if size == -1:
```
cwegrzyn, line 39: Hmm... Not understanding this change. From the language quoted above, seems like this should stay as `size < 0`? I can't imagine it comes up often, what kind of animal passes in size=-2, but specifications 🤷‍♂️ 

```suggestion
        if size < 0:
```

records_mover/utils/concat_files.py
```diff
@@ -18,10 +20,23 @@ def readable(self) -> bool:
         return True
 
     def readall(self) -> bytes:
-        return b''.join(f.read(-1) for f in self._files)
+        out = b''
```
cwegrzyn, line 23: You can take [this particular write up i found](https://www.guyrutenberg.com/2020/04/04/fast-bytes-concatenation-in-python/) with a grain of salt, but it matches my general understanding that concatenating plain strings can be slow. Swapping in a `bytearray` should behave the same with better performance. I guess you might need to cast it to a bytes at the very end to achieve the right types. Anyway, may not make a difference here, but if we end up reading a lot of big files, possible that it could.

records_mover/utils/concat_files.py
```diff
@@ -7,6 +7,8 @@ class ConcatFiles(io.RawIOBase):
 
     def __init__(self, files: List[IO[bytes]]) -> None:
         self._files = files
+        self._tell = 0
+        self._read_started = False
```
cwegrzyn, line 11: This line looks unused

```suggestion
```

cwegrzyn (approved):
A few minor things, but this looks generally good to me! :shipit:

records_mover/utils/concat_files.py
vinceatbluelabs: Ah, thanks - I hadn't seen that write-up!  

My expectation would be that the only thing that might call readall() would be Pandas, which is already pretty doomed by the time it's dealing with files large enough to split up.  Still makes sense to increase the threshold of doom, though!

PR #111, opened 2020-10-06
Prefer file extension to dialect compression defaults in targets

vinceatbluelabs:
A number of folks have been confused when they export files using something like `mvrec table2file --target.no_escape  <db details> myfile.csv` and find out that that the resulting file is GZIP compressed.

This change configures the system to use the following effective priority preference when determining what compression to use:

1. Explicit hint direction from the user -- i.e., telling us `--target.compression GZIP` implies GZIP compression.
2. Inference based on the filename - i.e., `myfile.csv` implies no compression.
3. Default for the specific dialect chosen -- i.e., `--target.dialect vertica` implies no compression.
4. Default for the default dialect - the bluelabs dialect implies GZIP compression.


tests/unit/records/targets/test_data_url.py
```diff
@@ -3,16 +3,19 @@
 from records_mover.records.targets.data_url import DataUrlTarget
 from records_mover.records.records_directory import RecordsDirectory
 from records_mover.records.base_records_format import BaseRecordsFormat
+from records_mover.records.records_format import DelimitedRecordsFormat
```
bluelabsbutler: F401 'records_mover.records.records_format.DelimitedRecordsFormat' imported but unused

records_mover/records/records_format.py
```diff
@@ -60,7 +60,7 @@ def __init__(self,
         """
         self.format_type = 'delimited'
         self.variant = variant
-        self._custom_hints = hints
+        self.custom_hints = hints
```
vinceatbluelabs, line 63: This isn't documented (and thus public for outside code to use), but it is now used by some of the target code.

records_mover/records/targets/data_url.py
```diff
@@ -22,14 +22,18 @@ def __init__(self,
                  output_loc: BaseFileUrl,
                  records_format: Optional[BaseRecordsFormat]) -> None:
         if records_format is None:
```
vinceatbluelabs, line 24: Since if any hint was provided in, a RecordsFormat object was created, we would have skipped the inference entirely.

records_mover/records/targets/data_url.py
```diff
@@ -22,14 +22,18 @@ def __init__(self,
...
-            # if we have been given free rein on format, don't
-            # surprise the user by writing a compression they're not
-            # expecting.
+            records_format = DelimitedRecordsFormat()
+
+        if (isinstance(records_format, DelimitedRecordsFormat) and
+           'compression' not in records_format.custom_hints):
```
vinceatbluelabs, line 28: Now so long as the user didn't specifically set a hint, we'll do inference on the file extension.

tests/unit/records/test_jobs.py
```diff
@@ -24,13 +26,12 @@ def mysource(a: int,
             self.assertEqual(google_cloud_creds, mock_session.creds.google_sheets.return_value)
             self.assertEqual(schema_name, 'myschema')
             self.assertEqual(spectrum_base_url, 'spectrumschema')
-            yield mock_source
+            return mock_source
 
-        @contextmanager
```
vinceatbluelabs, line 29: Whoops!  This never got fixed when sources/targets were made to be eager rather than lazy.  As a result, the assertions were never firing.

records_mover/records/records_format.py
```diff
@@ -73,14 +73,14 @@ def __eq__(self, other: object) -> bool:
...
-        input_hints = dict(self.hints)  # make copy
+        input_hints = dict(self.custom_hints)  # make copy
         input_hints.update(new_hints)
         return DelimitedRecordsFormat(variant=self.variant,
                                       hints=input_hints)  # type: ignore
 
     def alter_variant(self, variant: str) -> 'DelimitedRecordsFormat':
         return DelimitedRecordsFormat(variant=variant,
-                                      hints=self.hints)  # type: ignore
+                                      hints=self.custom_hints)
```
vinceatbluelabs, line 83: To make this work, we need to maintain a history of which hints were ever provided specifically.  Before, this history was being lost.  Indeed, this resulted in some unfortunate behavior.  For instance:

```sh
mvrec table2file --target.dialect csv --target.no_compression  ...
```

resulted in no compression, whereas:

```sh
mvrec table2file --target.no_compression --target.dialect csv ...
```

resulted in GZIP compression.

cwegrzyn (approved):
Makes sense to me-- the hints derived from the variants are not explicit instructions that we should preserve, whereas user-provided hints are. :shipit:

PR #112, opened 2020-10-06
Add Python 3.9 testing and local development, upgrade mypy for Python 3.9 compatibility

vinceatbluelabs:
Python 3.9 is out, and Records Mover seems to work with it!  I hit one minor glitch where the version of Mypy being used seemed to be putting out some bogus issues around BinaryIO vs Bytes[IO] when run under Python 3.9.  A quick Mypy upgrade seemed to fix those, so I bundled that in with this PR.


.circleci/config.yml
```diff
@@ -37,6 +37,13 @@ commands:
             else
               python -m venv venv
               . venv/bin/activate
+              # wheel is in requirements.txt, but without doing this first, the
+              # below pip3 install is very slow, with lots of warnings like:
+              #
+              #   Using legacy 'setup.py install' for Flask-JWT-Extended, since package 'wheel' is not installed.
+              #
+              # afterwards, it is merely pretty slow!
+              # pip install --upgrade wheel  # TODO see if this is still needed
```
bluelabsbutler, line 46: Uncompleted punchlist item detected--consider resolving or moving this to your issue tracker

.circleci/config.yml
```diff
@@ -50,7 +50,7 @@ commands:
               fi
             fi
       - save_cache:
-          key: deps-v1-<<parameters.python_version>>-<<parameters.pandas_version>>-<<parameters.extras>>-<<parameters.include_dev_dependencies>>-{{ .Branch }}-{{ checksum "requirements.txt" }}-{{ checksum "setup.py" }}
+          key: deps-v2-<<parameters.python_version>>-<<parameters.pandas_version>>-<<parameters.extras>>-<<parameters.include_dev_dependencies>>-{{ .Branch }}-{{ checksum "requirements.txt" }}-{{ checksum "setup.py" }}
```
vinceatbluelabs, line 53: CircleCI keeps caches immutably - you can't flush them.  This versioning lets me flush the cache when I need to test something.

records_mover/records/delimited/types.py
```diff
@@ -57,7 +57,7 @@
                    "datetimeformattz",
                    "datetimeformat"]
 
-HINT_NAMES: List[HintName] = list(get_args(HintName))  # type: ignore
+HINT_NAMES: List[HintName] = list(get_args(HintName))
```
vinceatbluelabs, line 60: Looks like the new mypy is better at understanding `get_args()`!

records_mover/utils/json_schema.py
```diff
@@ -21,7 +21,9 @@
 
 def is_iterable_type(python_type: PythonType) -> bool:
     return get_origin(python_type) in [
-        collections.abc.Iterable,
+        # mypy reports error: Module has no attribute "abc", which is not true:
+        # https://docs.python.org/3/library/collections.abc.html
+        collections.abc.Iterable,  # type: ignore
```
vinceatbluelabs, line 26: ...but worse at understanding the collections package?

tests/unit/records/test_hints.py
```diff
@@ -17,7 +18,7 @@ def test_sniff_hints_from_fileobjs(self,
                                        mock_csv) -> None:
         mock_fileobj = MagicMock(name='fileobj')
         mock_fileobj.closed = False
-        mock_fileobjs = [mock_fileobj]
+        mock_fileobjs: List[IO[bytes]] = [mock_fileobj]
```
vinceatbluelabs, line 21: Mypy was smart enough to complain that I was passing mock objects to a typed method.  Looks like it's happy if I tell it what type the value is pretending to be, though!

.circleci/config.yml
```diff
@@ -408,6 +408,16 @@ workflows:
...
+          requires:
+            - redshift-s3-itest
+          filters:
+            tags:
+              only: /v\d+\.\d+\.\d+(-[\w]+)?/
+      - test:
+          name: test-3.9
+          extras: '[unittest,typecheck]'
+          python_version: "3.9"
+          pandas_version: ">=1"
```
cwegrzyn, line 420: Looks like this one doesn't require `redshift-s3-itest` -- is that intentional for a single, recent python?

cwegrzyn (approved):
LGTM! I had one question about the circleci config-- but I think it was intentional... so :shipit:

.circleci/config.yml
vinceatbluelabs: Yep, that's as expected.  The build starts out with a single unit test task, then a single Redshift integration test, then branches out into everything else.  This PR introduces test-3.9 and relegates test-3.8 to the back of the pack:

image

PR #114, opened 2020-10-27
Load via INSERT on Redshift when scratch bucket not available

vinceatbluelabs:
This PR teaches Records Mover that it can use INSERT statements to import to Redshift when nothing else is possible.  INSERT doesn't involve AWS credentials and a temporary S3 bucket.

This is sort of useful on its own, but primarily it sets me up in the next PR (#113) to do something similar for BigQuery, so folks don't *need* to configure a BigQuery Google Cloud Storage bucket just to be able to do imports like they're able to do today.

PR #115, opened 2020-10-27
Retire SupportsMoveFromTempLocAfterFillingIt from DataUrlTarget

vinceatbluelabs:
It's not needed anymore - the integration tests still pass, and it doesn't make any sense on its face.

I ran across the need to change this interface (see #114) and noticed that it was being used in a nonsensical way here.


records_mover/records/targets/data_url.py
```diff
@@ -63,25 +60,6 @@ def move_from_records_directory(self,
...
-                                      f"{records_source} and {self}")
-        with self.output_loc.temporary_directory() as temp_loc:
-            directory = RecordsDirectory(records_loc=temp_loc)
-            records_source.\
-                move_to_records_directory(directory,
-                                          records_format=records_format,
-                                          processing_instructions=pis)
-            return self.move_from_records_directory(directory,
-                                                    processing_instructions)
-
```
vinceatbluelabs, line 84: This function said "Move something to this filesystem by first creating a temporary directory ON THIS SAME FILESYSTEM and then copying from there".  Kind of silly, since you can also just copy directly to the intended destination (see the 

The purpose of `move_from_temp_loc_after_filling_it()` is for systems like Redshift where the fastest way to copy things in is from an S3 bucket.  In those cases, even if the source isn't an S3 bucket, we want to copy to a temporary S3 bucket before loading.

I'm guessing this function in this file was implemented at some point of time in the past in order to work around some wart or another in the `move()` algorithm.  The `move()` algorithm uses these Python interfaces to figure out a way to get data from the source to the target in a multiple dispatch way.  

There's been a lot of incremental development and refactoring among all of these things, and evolutionary dead-ends are not unknown :)

records_mover/records/targets/data_url.py
```diff
@@ -16,7 +14,6 @@
 
 
 class DataUrlTarget(SupportsMoveFromDataframes,
```
vinceatbluelabs, line 16: Targets in Records Mover represent the second part in a move (e.g., in a Redshift to S3 copy, S3 is the target and Redshift is the source).

DataUrlTarget represents any kind of filesystem-like thing - currently it works with `file://`, `gs://` and `s3://` URLs, but will probably get `sftp://` support at some point as well.

vinceatbluelabs:
Hi @crvena-sonja!  Just pinging on this review.  This is review is mainly for a second developer to have a little more familiarity with some Records Mover internals.

crvena-sonja (approved):
☠️

PR #116, opened 2020-10-29
Clarify semantics of can_load_direct()/can_move_from_this_format() and fix incorrect implementation

vinceatbluelabs:
The [move()](https://github.com/bluelabsio/records-mover/blob/master/records_mover/records/mover.py#L18) algorithm is a little subtle, and I found myself a little confused while working through solving a problem with it recently.

Part of my confusion came from some unclear naming/semantics around a couple of functions  there - this PR cleans that up in a way that seems more clear to me!


records_mover/records/mover.py
```diff
@@ -62,7 +62,7 @@ def move(records_source: RecordsSource,
     # being called.
     if (isinstance(records_source, sources_base.SupportsRecordsDirectory) and
         isinstance(records_target, SupportsMoveFromRecordsDirectory) and
-       records_target.can_load_direct(records_source.records_directory().loc.scheme) and
+       records_target.can_load_directly_from_this_scheme(records_source.records_directory().loc.scheme) and
```
bluelabsbutler, line 65: E501 line too long (107 > 100 characters)

records_mover/records/targets/data_url.py
```diff
@@ -48,8 +48,11 @@ def move_from_dataframes_source(self,
                 move_from_dataframes_source(dfs_source=dfs_source,
                                             processing_instructions=processing_instructions)
 
-    def can_load_direct(self, scheme: str) -> bool:
-        return scheme == self.output_loc.scheme
+    def can_move_directly_from_scheme(self, scheme: str) -> bool:
+        # Currently all means of copying between different schemes
+        # involve streaming data down to Records Mover from the source
+        # and then back up to the target.
+        return False
```
vinceatbluelabs, line 55: The integration tests pass fine here, so I don't think the implementation here has any effect currently - but I wanted it to be correct in case the move() logic ends up depending on it, and also so that the intended semantics are clear.

Note that arguably this could be set to 'true' for file -> file URL operations, since nothing leaves the box.  I might extend it to do that someday when that's relevant to optimizing a move() path.

crvena-sonja: I honestly prefer the empty state to be False rather than True, mainly to avoid side effects.... 👍

crvena-sonja (approved):
Thank you for the helpful comments here. LGTM. :shipit:

PR #117, opened 2020-11-05
Do slow redshift unload via SELECT when bucket unload not available

vinceatbluelabs:
This allows Redshift to export data even when we don't have a scratch bucket to copy it to first.

The particular method is similar to the fix done for importing data (#114) - adding some predicates to the Source/Target and the underlying DBDriver interfaces so that the move() algorithm can figure out whether for a given URL scheme it can do a bulk export or whether the move() algorithm should do things via SELECT.


records_mover/db/unloader.py
```diff
@@ -25,10 +25,15 @@ def unload(self,
     def known_supported_records_formats_for_unload(self) -> List[BaseRecordsFormat]:
         ...
 
+    # TODO: Drop this in name in other PR
```
bluelabsbutler, line 28: Uncompleted punchlist item detected--consider resolving or moving this to your issue tracker

records_mover/records/sources/table.py
```diff
@@ -53,6 +53,13 @@ def can_move_to_this_format(self,
             return False
         return unloader.can_unload_this_format(target_records_format)
 
+    # TODO: above can_move_to_this_format vs can_move_to_scheme - do I need to modify other PR?
```
bluelabsbutler, line 56: Uncompleted punchlist item detected--consider resolving or moving this to your issue tracker

metrics/coverage_high_water_mark
```diff
@@ -1 +1 @@
-93.6900
+93.7300
```
vinceatbluelabs, line 1: I tested the bulk of the new code added, and threw in some unrelated tests as well to get this number up.

records_mover/db/loader.py
```diff
@@ -42,6 +42,9 @@ def can_load_this_format(self, source_records_format: BaseRecordsFormat) -> bool
         method"""
         ...
 
+    def temporary_loadable_directory_scheme(self) -> str:
+        return 'file'
```
vinceatbluelabs, line 46: As a refresher:

* DBDriver abstract classes: This is a wrapper on top of SQLAlchemy's database dialects that describe how to load/unload data and also provide some extra information about things like the numeric ranges of types.
* Loader/Unloader abstract classes: These are the interfaces extended for each database to show how to do bulk imports/exports.
* source/target: sources and targets are things like files on disk, directories in S3, tables in databases, etc - you move data with Records Mover from a source to a target.
* Source/Target abstract classes: these are abstract interfaces that show what capabilities a given source or target have.  The table source and table target of course use the DBDriver interface extensively to know what capabilities a specific database have.
* move() algorithm - uses Source and Target objects and inspect to find out which particular Source/Target abstract classes are implemented to figure out the fastest way to move data from a source to a target.

So this is an additional exported bit of info on the Loader interface that helps us know that if we had to create a temporary directory to load from, where it would be created.  We'll see where that's used later on in the PR here.

records_mover/db/postgres/unloader.py
```diff
@@ -101,6 +101,11 @@ def known_supported_records_formats_for_unload(self) -> List[BaseRecordsFormat]:
                                    }),
         ]
 
+    def can_unload_to_scheme(self, scheme: str) -> bool:
+        # Unloading is done via streams, so it is scheme-independent
+        # and requires no scratch buckets.
+        return True
```
vinceatbluelabs, line 107: In one movement scenario, we ask a source database to export into a records directory (basically a pile of CSVs with a JSON file for metadata).  This is how we can find out if the source database can export into the right kind of place for the target database (or whatever) to import.

records_mover/db/loader.py
```diff
@@ -42,6 +42,9 @@ def can_load_this_format(self, source_records_format: BaseRecordsFormat) -> bool
         method"""
         ...
 
+    def temporary_loadable_directory_scheme(self) -> str:
```
vinceatbluelabs, line 45: - [ ] TODO: Document this method

records_mover/db/unloader.py
```diff
@@ -29,6 +29,10 @@ def known_supported_records_formats_for_unload(self) -> List[BaseRecordsFormat]:
     def can_unload_this_format(self, target_records_format: BaseRecordsFormat) -> bool:
         ...
 
+    @abstractmethod
+    def can_unload_to_scheme(self, scheme: str) -> bool:
```
vinceatbluelabs, line 33: - [ ] TODO: Document this method

records_mover/db/vertica/unloader.py
```diff
@@ -47,7 +47,7 @@ def unload(self,
                table: str,
                unload_plan: RecordsUnloadPlan,
                directory: RecordsDirectory) -> Optional[int]:
-        if not self.s3_available():
+        if not self.s3_export_available():
```
vinceatbluelabs, line 50: I'm separating out the concepts of "does the database have the right add-ons installed to export to S3" and "do we have a temporary S3 bucket available".

Even if we don't have a *temporary* S3 bucket available, we may be able to export straight to the destination S3 bucket!

records_mover/db/vertica/unloader.py
```diff
@@ -119,15 +119,14 @@ def unload_to_s3_directory(self,
         return export_count
 
     def known_supported_records_formats_for_unload(self) -> List[BaseRecordsFormat]:
-        if self.s3_available():
-            return [DelimitedRecordsFormat(variant='vertica')]
-        else:
-            return []
+        return [DelimitedRecordsFormat(variant='vertica')]
```
vinceatbluelabs, line 122: This was an existing kludge to try to accomplish some of the "negotiation" happening with this PR.  Now that we have it more fully baked at the move() algorithm level, we can tear down this hack, as it's no longer needed.

records_mover/records/mover.py
```diff
@@ -126,6 +127,8 @@ def move(records_source: RecordsSource,
     elif (isinstance(records_source, SupportsMoveToRecordsDirectory) and
           isinstance(records_target, MightSupportMoveFromTempLocAfterFillingIt) and
           records_source.has_compatible_format(records_target) and
+          records_source.
+            can_move_to_scheme(records_target.temporary_loadable_directory_scheme()) and
```
vinceatbluelabs, line 131: These three lines added lines are the meat of this PR.  They result in us not trying to do bulk exports when we'd need a temporary bucket and don't have one.  As a result, we move on and try a less efficient approach rather than dying with an exception.

records_mover/records/sources/base.py
```diff
@@ -96,6 +96,16 @@ def records_directory(self) -> RecordsDirectory:
...
 class SupportsMoveToRecordsDirectory(NegotiatesRecordsFormat, metaclass=ABCMeta):
+    @abstractmethod
+    def can_move_to_scheme(self, scheme: str) -> bool:
+        """If true is returned, the given scheme is a compatible place where
+        the unload will be done.  Note that this may include streaming
+        data down to Records Mover byte by byte--which can be
+        expensive when data is large and/or network bandwidth is
+        limited.
+        """
+        pass
```
vinceatbluelabs, line 107: This is one of the abstract classes that show what capabilities a Source has.

records_mover/records/targets/table/target.py
```diff
@@ -116,6 +116,14 @@ def can_move_from_temp_loc_after_filling_it(self) -> bool:
             return False
         return loader.has_temporary_loadable_directory_loc()
 
+    def temporary_loadable_directory_scheme(self) -> str:
+        driver = self.db_driver(self.db_engine)
+        loader = driver.loader()
+        if loader is None:
+            raise TypeError("Please check can_move_from_temp_loc_after_filling_it() "
+                            "before calling this")
```
vinceatbluelabs, line 124: This isn't the ideal from a type safety perspective, but I looked at a few alternatives and couldn't come up with a better approach without making the move() method much uglier.

records_mover/records/sources/table.py
```diff
@@ -53,6 +53,12 @@ def can_move_to_this_format(self,
             return False
         return unloader.can_unload_this_format(target_records_format)
 
+    def can_move_to_scheme(self, scheme: str) -> bool:
+        unloader = self.driver.unloader()
+        if unloader is None:
+            return False
```
vinceatbluelabs, line 59:
```suggestion
            # bulk export is not provided by this database
            return False
```

records_mover/db/loader.py
```diff
@@ -42,12 +42,22 @@ def can_load_this_format(self, source_records_format: BaseRecordsFormat) -> bool
...
     @contextmanager
     def temporary_loadable_directory_loc(self) -> Iterator[BaseDirectoryUrl]:
+        """Provide a temporary directory which can be used for bulk import to
+        this database and clean it up when done"""
         with TemporaryDirectory(prefix='temporary_loadable_directory_loc') as dirname:
             yield FilesystemDirectoryUrl(dirname)
 
     def has_temporary_loadable_directory_loc(self) -> bool:
+        """Returns True if a temporary directory can be provided by
+        temporary_loadable_directory_loc()"""
```
vinceatbluelabs, line 60: Be more consistent with documentation in this class.

tests/unit/db/redshift/test_unloader.py
```diff
@@ -33,3 +33,48 @@ def test_can_unload_this_format_true(self,
...
+        mock_db = Mock(name='db')
+        mock_table = Mock(name='table')
+
+        redshift_unloader =\
+            RedshiftUnloader(db=mock_db,
+                             table=mock_table,
+                             s3_temp_base_loc=None)
+        self.assertTrue(redshift_unloader.can_unload_to_scheme('s3'))
+
+    def test_can_unload_to_scheme_file_with_temp_bucket_True(self):
```
crvena-sonja, line 47: the most pedantic comment, but we're here.... True is capitalized in the function name but not in others.

records_mover/records/targets/table/target.py
crvena-sonja: With the understanding that this not shown to the External API, and mostly as a signal to fix logic elsewhere, I see no reason to belabor this point.

crvena-sonja (approved):
One comment about capitalizing True in a function name, but otherwise LGTM.

tests/unit/db/redshift/test_unloader.py
```diff
@@ -33,3 +33,48 @@ def test_can_unload_this_format_true(self,
...
+        mock_db = Mock(name='db')
+        mock_table = Mock(name='table')
+
+        redshift_unloader =\
+            RedshiftUnloader(db=mock_db,
+                             table=mock_table,
+                             s3_temp_base_loc=None)
+        self.assertFalse(redshift_unloader.can_unload_to_scheme('file'))
+
+    def test_can_unload_to_scheme_file_with_temp_bucket_true(self):
```
bluelabsbutler, line 57: F811 redefinition of unused 'test_can_unload_to_scheme_file_with_temp_bucket_true' from line 47

PR #120, opened 2020-11-08
Allow environment-based configuration of GCS creds

vinceatbluelabs:
Unlike AWS, Google Cloud Platform doesn't provide the ability to configure cloud credentials directly via environment variables.  The best they do is to provide an environment variable that points to a file.

This limitation provides some friction for Records Mover when used in various different environments, so as a workaround, I allow users to set the GCP_SERVICE_ACCOUNT_JSON_BASE64 variable and then call into GCP's API.

This was set up originally for BigQuery, but it appears I didn't extend that out to also work with Google Cloud Storage or Google Sheets.

That's what this PR fixes.

An immediate place where this is useful (and the forcing function for this PR) is that we can set env variables in CircleCI to configure credentials without having to write shell script glue to take that environment variable, write it to a file, and then point to it with another env variable.


records_mover/creds/base_creds.py
```diff
@@ -13,6 +13,12 @@
 
 logger = logging.getLogger(__name__)
 
+_GSHEETS_SCOPES = ('https://www.googleapis.com/auth/spreadsheets',)
+
+_GCS_SCOPES = ('https://www.googleapis.com/auth/devstorage.full_control',
+               'https://www.googleapis.com/auth/devstorage.read_only',
+               'https://www.googleapis.com/auth/devstorage.read_write')
```
vinceatbluelabs, line 20: Another more minor thing this PR fixes is that it limits the scope of the creds we pull when we pull from the default GCP creds store.

(The default GCP creds store is what gets populated when you either set the env variable that points to service account JSON, or you log in with `gcloud auth application-default login`)

records_mover/creds/base_creds.py
```diff
@@ -94,19 +98,23 @@ def default_boto3_session(self) -> Optional['boto3.session.Session']:
             self.__default_boto3_session = self.boto3_session(self._default_aws_creds_name)
         return self.__default_boto3_session
 
+    def _gcp_creds_of_last_resort(self,
+                                  scopes: Iterable[str]) ->\
+            Optional['google.auth.credentials.Credentials']:
+        import google.auth
+        credentials, project = google.auth.default(scopes=scopes)
+        return credentials
```
vinceatbluelabs, line 106: By extracting out this logic to a function, subclasses (like `CredsViaEnv`) can overwrite it and pull credentials from things like environment variables if more specific creds haven't been pointed to with a cred name.

records_mover/creds/base_creds.py
```diff
@@ -94,19 +98,23 @@ def default_boto3_session(self) -> Optional['boto3.session.Session']:
...
         if self.__default_gcs_creds is not PleaseInfer.token:
             return self.__default_gcs_creds
 
         try:
             import google.auth.exceptions
             if self._default_gcp_creds_name is None:
-                import google.auth
-                credentials, project = google.auth.default()
-                self.__default_gcs_creds = credentials
+                self.__default_gcs_creds = self._gcp_creds_of_last_resort(scopes=_GCS_SCOPES)
```
vinceatbluelabs, line 115: Note that now I'm limiting this to specific scopes.

records_mover/creds/creds_via_env.py
```diff
@@ -13,19 +13,41 @@
...
+            Optional['google.auth.credentials.Credentials']:
+        if 'GCP_SERVICE_ACCOUNT_JSON_BASE64' not in os.environ:
+            return None
         import google.oauth2.service_account
-        notes_json_base64 = os.environ['GCP_SERVICE_ACCOUNT_JSON_BASE64']
-        notes_json = base64.b64decode(notes_json_base64.encode('utf-8'))
-        cred_details = json.loads(notes_json)
+        json_base64 = os.environ['GCP_SERVICE_ACCOUNT_JSON_BASE64']
+        json_str = base64.b64decode(json_base64.encode('utf-8'))
+        cred_details = json.loads(json_str)
```
vinceatbluelabs, line 24: The weird names here were a copy-and-pasto from the CredsViaLastPass class.

records_mover/creds/creds_via_env.py
```diff
@@ -13,19 +13,41 @@
...
 
+    def _gcp_creds(self, gcp_creds_name: str,
+                   scopes: Iterable[str]) -> 'google.auth.credentials.Credentials':
+        # No current way to differentiate between different creds;
+        # just on env variable recognized.
+        creds_from_env = self._gcp_creds_from_env(scopes)
+        if creds_from_env is None:
+            raise KeyError('Please set GCP_SERVICE_ACCOUNT_JSON_BASE64 to configure Google '
+                           'Cloud Platform credentials')
+        return creds_from_env
```
vinceatbluelabs, line 37: Note that in practice our internal use on laptops uses `CredsViaLastPass`, but this class gets used in places like JupyterHub and CircleCI.

types/stubs/google/auth/__init__.pyi
```diff
@@ -1,6 +1,6 @@
 from google.auth.credentials import Credentials
-from typing import Tuple
+from typing import Tuple, Iterable
 
 
-def default() -> Tuple[Credentials, str]:
+def default(scopes: Iterable[str]) -> Tuple[Credentials, str]:
```
vinceatbluelabs, line 5: This line updates our internal typestubs to include the additional supported parameter to this method.

records_mover/creds/creds_via_env.py
```diff
@@ -13,19 +13,41 @@
...
+
     def db_facts(self, db_creds_name: str) -> DBFacts:
         return db(db_creds_name.split('-'))
 
+    def _gcp_creds_of_last_resort(self,
```
crvena-sonja, line 42: 😆

crvena-sonja (approved):
LGTM. Like the security best-practices here, too.

PR #123, opened 2020-11-09
Drop dead code

vinceatbluelabs:


records_mover/creds/creds_via_env.py
```diff
@@ -25,11 +25,3 @@ def _gcp_creds(self, gcp_creds_name: str,
...
     def db_facts(self, db_creds_name: str) -> DBFacts:
         return db(db_creds_name.split('-'))
-
-    def boto3_session(self, aws_creds_name: str) -> 'boto3.session.Session':
-        import boto3
-
-        if aws_creds_name is None:
-            return boto3.session.Session()
-        else:
-            raise NotImplementedError
```
vinceatbluelabs, line 35: We verify via mypy that `aws_creds_name` is never None - so the only thing this will ever do is `raise NotImplementedError`.  Since that's what the superclass does anyway, we can drop this.

I believe this code was relevant at one point, but then we created a separate `.default_boto3_session()` method.

crvena-sonja (approved):
🔥

PR #125, opened 2020-11-09
Rename records_mover.records.types

vinceatbluelabs:
Renamed to records_mover.records.records_types in order to avoid an issue in mypy where in local IDE verification mode, it doesn't like modules named "types" and as a result you don't get type error messages in IDEs.

(it works fine in CircleCI and on the command line regardless)


crvena-sonja (approved):
:shipit:

PR #126, opened 2020-11-09
Retry on more Google exceptions

vinceatbluelabs:
We have a decorator which does retries when Google API operations send us a rate limit error.  It avoids having to manually restart integration tests in CI (where a lot of stuff in parallel can trigger these rate limits).

See https://app.circleci.com/pipelines/github/bluelabsio/records-mover/1314/workflows/4913c847-9089-461f-838d-3955b27032b4/jobs/11584 for an example failure during integration testing in CI - this wasn't covered by the existing decorator.


crvena-sonja (approved):
🙄 string matching ftw.

PR #127, opened 2020-11-10
Add keepalive config for Redshift

vinceatbluelabs:


vinceatbluelabs:
Appears that large Redshift exports are reliable after #128 and #133!

PR #113, opened 2020-10-26
BigQuery import from GCS buckets

vinceatbluelabs:
Right now we only have code to load data into BigQuery by streaming data over the network to it.  This PR teaches it as well to be able to point to existing data in a GCS bucket.

This is important when data gets large - there are ways to move mass amount of data into a GCS bucket (e.g., GCP Cloud Data Transfer) that accomplish the goal much faster than trying to stream data down to your laptop and back up again.  BigQuery is able to read from a bucket in parallel and over a much faster network in order to load data.

One major limitation remaining: all of the methods that Records Mover knows about to get data *into* that GCS bucket are inherently slow as they rely on streaming data down to Records Mover and then back up to the GCS bucket.  In practice for large datasets (>100MB) folks will want to first get data into a GCS bucket and then use Records Mover.

Watch this space.

Depends on https://github.com/bluelabsio/records-mover/pull/120 and set to merge into it for clarity of diff.

PR #134, opened 2020-11-14
Bug fix: Fall back to loading via network stream if GCS bucket not available

vinceatbluelabs:
BigQuery is the first database supported by Records Mover that supports bulk loading via both a bucket-based mechanism as well as via streams ("fileobjs" in Python lingo).

It looks like the negotiation logic in Records Mover is not smart enough to fall back to loading via streams if loading via a bucket is not available.  As a result, #113 resulted in existing behavior getting slower - folks who don't have a scratch bucket configured and aren't loading from a GCS bucket already will see loads happen via INSERT instead of via a network stream of a CSV.  This is generally known to be much slower for databases.

This PR teaches the negotiation logic to fall back smartly from bucket -> stream -> INSERT.


records_mover/records/targets/table/move_from_temp_loc_after_filling_it.py
```diff
@@ -7,7 +7,7 @@
 from records_mover.records.sources import SupportsMoveToRecordsDirectory
 from records_mover.records.targets.table.base import BaseTableMoveAlgorithm
 import logging
-from typing import Iterator, TYPE_CHECKING
+from typing import Iterator, Optional, TYPE_CHECKING
```
bluelabsbutler, line 10: F401 'typing.Optional' imported but unused

records_mover/db/bigquery/loader.py
```diff
@@ -32,9 +32,13 @@ def __init__(self,
         self.url_resolver = url_resolver
         self.gcs_temp_base_loc = gcs_temp_base_loc
 
+    def best_scheme_to_load_from(self) -> str:
+        return 'gs'
+
     @contextmanager
     def temporary_gcs_directory_loc(self) -> Iterator[BaseDirectoryUrl]:
         if self.gcs_temp_base_loc is None:
+            # TODO: Can this yield None succesfully?  Would make logic more typesafe.
```
bluelabsbutler, line 41: Uncompleted punchlist item detected--consider resolving or moving this to your issue tracker

records_mover/db/bigquery/loader.py
```diff
@@ -102,15 +105,21 @@ def load(self,
...
+                    gcs_directory = directory.copy_to(temp_gcs_loc)
+                    return self.load(schema=schema,
+                                     table=table,
+                                     load_plan=load_plan,
+                                     directory=gcs_directory)
+            else:
+                return self.load_from_records_directory_via_fileobj(schema=schema,
+                                                                    table=table,
+                                                                    load_plan=load_plan,
+                                                                    directory=directory)
```
vinceatbluelabs, line 122: If we don't have a temp bucket, load via stream!

records_mover/db/loader.py
```diff
@@ -80,21 +82,32 @@ def load_from_fileobj(self, schema: str, table: str,
...
-            return self.load_from_fileobj(schema,
-                                          table,
-                                          load_plan,
-                                          concatted_fileobj)
-        return None  # make mypy happy
+        # Implement load() in terms of load_from_fileobj() by default
+        return self.load_from_records_directory_via_fileobj(schema=schema,
+                                                            table=table,
+                                                            load_plan=load_plan,
+                                                            directory=directory)
```
vinceatbluelabs, line 113: This logic only worked for files which could be concatenated - and didn't have any safety checks to avoid sending the database invalid data files.  Basically, it worked for CSV files without headers, or single files.

The new logic simply loads multiple streams in a row, and relies on load_from_fileobj() having append semantics (I documented that requirement and verified the existing implementations comply).

records_mover/db/postgres/unloader.py
```diff
@@ -106,6 +110,11 @@ def can_unload_to_scheme(self, scheme: str) -> bool:
         # and requires no scratch buckets.
         return True
 
+    @contextmanager
+    def temporary_unloadable_directory_loc(self) -> Iterator[BaseDirectoryUrl]:
+        with TemporaryDirectory(prefix='temporary_unloadable_directory_loc') as dirname:
+            yield FilesystemDirectoryUrl(dirname)
```
vinceatbluelabs, line 116: If we're doing a table2table copy between, say, Redshift and BigQuery, and we don't have a temporary bucket on the BigQuery side, the move() logic needs to be able to pull a bucket directory that we can *un*load into.

records_mover/db/vertica/unloader.py
```diff
@@ -54,7 +54,7 @@ def unload(self,
             if directory.scheme == 's3':
                 return self.unload_to_s3_directory(schema, table, unload_plan, directory)
             else:
-                with self.temporary_loadable_directory_loc() as temp_s3_loc:
+                with self.temporary_unloadable_directory_loc() as temp_s3_loc:
```
vinceatbluelabs, line 57: This was a copy-and-pasto from way back.

records_mover/records/targets/table/target.py
```diff
@@ -118,6 +118,11 @@ def can_move_from_temp_loc_after_filling_it(self) -> bool:
         if loader is None:
             logger.warning(f"No loader configured for this database type ({self.db_engine.name})")
             return False
+        if self.can_move_from_fileobjs_source():
+            # Regardless we can certainly stream the data from wherever
+            # the source writes it
+            return True
```
vinceatbluelabs, line 124: This allows the move() algorithm to go into this case to begin with.

records_mover/db/bigquery/loader.py
```diff
@@ -102,15 +105,21 @@ def load(self,
              schema: str,
              table: str,
              load_plan: RecordsLoadPlan,
-             directory: RecordsDirectory) -> int:
+             directory: RecordsDirectory) -> Optional[int]:
```
vinceatbluelabs, line 108: We're now calling into a generic interface (`. load_from_records_directory_via_fileobj()` below which doesn't know if the implementation can actually return the number of rows loaded.  Rest assured that this one can (not that anything depends on that - this is bubbled up as Optional throughout).

PR #135, opened 2020-11-15
Fix dependencies for Homebrew processing

vinceatbluelabs:
The primary motivation for this one is to get `brew update-python-resources` to work.

`brew update-python-resources` is the modern version of [homebrew-pypi-poet](https://github.com/tdsmith/homebrew-pypi-poet), and uses [pipgrip](https://github.com/ddelange/pipgrip) for depenency processing.  `pipgrip`, in turn, "is a lightweight pip dependency resolver with deptree preview functionality based on the PubGrub algorithm, which is also used by poetry".

This also helps (but doesn't completely solve) getting Records Mover to install using the upcoming version of `pip`, which can be tested using `pip3 install --use-feature=2020-resolver`.  So far that still hangs in an apparently infinite loop while trying to resolve Records Mover dependencies.  Hopefully that will shake out as our dependencies start to test and push fixes out to be compatible with the new resolver algorithm.


setup.py
```diff
@@ -138,14 +138,30 @@ def initialize_options(self) -> None:
 ]
 
 db_dependencies = [
-    # sqlalchemy 1.3.16 seems to have (accidentally?) introduced
-    # a breaking change that affects sqlalchemy-redshift:
+    # Without this, pip 2.20.4 with "--use-feature=2020-resolver" will
```
bluelabsbutler, line 141: This file, one of the 3 largest in the project, increased in size to 373 lines.  The total size of those files is now 3415 lines (target: 1087).  Is this file complex enough to refactor?

bluelabsbutler, line 141: This file, one of the 3 largest in the project, increased in size to 373 lines.  The total size of those files is now 3415 lines (target: 1103).  Is this file complex enough to refactor?

vinceatbluelabs: Yeah, it did.  That's a bummer.  We've tried breaking this file into separate files before, and found that Python doesn't support that in all `pip install` scenarios.

PR #137, opened 2020-12-01
Add Avro support for Redshift import

vinceatbluelabs:
This PR allows Redshift to import from Avro, which clears the path for BigQuery -> Avro moves to happen via Avro files copied from GCS to S3 buckets.

The bummer of this is that Redshift doesn't support date/time types in Avro, which means a bunch more special cases for date/times in the tests.

This PR also adds Avro support in an extra place or two I missed in the last PR - revealing some things that might be good to get into component/integration tests in the future:
* On the command line `--format` argument.
* As a filename extension to parse and then infer a records format from.
* As a records format filename to parse in a Records Directory (e.g., `_format_avro`)


records_mover/db/redshift/records_copy.py
```diff
@@ -1,5 +1,6 @@
 from ...utils import quiet_remove
 from ...records.delimited import cant_handle_hint, ValidatedRecordsHints
+from records_mover.records.records_format import BaseRecordsFormat, DelimitedRecordsFormat, AvroRecordsFormat
```
bluelabsbutler, line 3: E501 line too long (109 > 100 characters)

records_mover/db/bigquery/unloader.py
```diff
@@ -33,7 +33,11 @@ def can_unload_format(self, target_records_format: BaseRecordsFormat) -> bool:
         return False
 
     def can_unload_to_scheme(self, scheme: str) -> bool:
-        return scheme == 'gs'
+        if scheme == 'gs':
+            return True
+        # Otherwise we'll need a temporary bucket configured for
+        # BigQuery to unload into
+        return self.gcs_temp_base_loc is not None
```
vinceatbluelabs, line 40: With this we can unload via a GCS scratch bucket.  I forgot this in the last PR!

records_mover/records/__init__.py
```diff
@@ -20,7 +21,9 @@
 from .records_types import RecordsFormatType, DelimitedVariant
 from .schema import RecordsSchema
 from .mover import move
-from .records_format import RecordsFormat, DelimitedRecordsFormat, ParquetRecordsFormat
+from .records_format import (
+    RecordsFormat, DelimitedRecordsFormat, ParquetRecordsFormat, AvroRecordsFormat
+)
```
vinceatbluelabs, line 26: Let's make this a public interface similar to DelimitedRecordsFormat/ParquetRecordsFormat

records_mover/records/records_format.py
```diff
@@ -43,6 +48,11 @@ def __repr__(self) -> str:
     def generate_filename(self, basename: str) -> str:
         return f"{basename}.parquet"
 
+    def __eq__(self, other: object) -> bool:
+        if isinstance(other, ParquetRecordsFormat):
+            return True
+        return False
+
```
vinceatbluelabs, line 55: The above `__eq__()` methods are primarily convenience things; they make the unit tests nicer when comparing.  We'll need to make them more complex once these records formats have hints associated with them (e.g., compression type).

records_mover/records/schema/schema/__init__.py
```diff
@@ -15,6 +15,7 @@
     from mypy_extensions import TypedDict
 
     from ..field import FieldDict
+    from ..field.field_types import FieldType
```
bluelabsbutler, line 18: F401 '..field.field_types.FieldType' imported but unused

records_mover/db/bigquery/bigquery_db_driver.py
```diff
@@ -135,6 +134,6 @@ def tweak_records_schema_after_unload(self,
             # DATETIME field."
             #
             # BigQuery exports this as an Avro string type
-            return records_schema.convert_datetimes_to_string()
+            return records_schema.cast_field_types({'datetime': 'string'})
```
vinceatbluelabs, line 137: Since we seem to be using it more and more, let's at least generalize this function a bit.

records_mover/db/redshift/loader.py
```diff
@@ -114,19 +112,20 @@ def can_load_this_format(self, source_records_format: BaseRecordsFormat) -> bool
...
+                                  records_format,
                                   processing_instructions.fail_if_cant_handle_hint,
                                   processing_instructions.fail_if_row_invalid,
                                   processing_instructions.max_failure_rows)
-            complain_on_unhandled_hints(processing_instructions.fail_if_dont_understand,
-                                        unhandled_hints, load_plan.records_format.hints)
+            if isinstance(records_format, DelimitedRecordsFormat):
+                complain_on_unhandled_hints(processing_instructions.fail_if_dont_understand,
+                                            unhandled_hints,
+                                            records_format.hints)
```
vinceatbluelabs, line 128: The changes above are to stop assuming we're dealing with a DelimitedRecordsFormat object and let the other types through.

records_mover/db/redshift/loader.py
```diff
@@ -158,6 +157,7 @@ def known_supported_records_formats_for_load(self) -> List[BaseRecordsFormat]:
             }),
             # Supports newlines in strings, but not empty strings.
             DelimitedRecordsFormat(variant='bluelabs'),
+            AvroRecordsFormat(),
```
vinceatbluelabs, line 160: Let's set Avro as the least preferred format for now.  Maybe in the future Records Mover could be smart enough to fix up the columns after loading the file?

records_mover/records/schema/field/__init__.py
```diff
@@ -313,40 +313,9 @@ def from_data(name: str, data: 'FieldDict') -> 'RecordsSchemaField':
...
-                statistics = statistics.cast('string')
-        return RecordsSchemaField(name=self.name,
-                                  field_type=field_type,
-                                  constraints=constraints,
-                                  statistics=statistics,
-                                  representations=self.representations)
-
     def cast(self, field_type: 'FieldType') -> 'RecordsSchemaField':
+        if self.field_type == field_type:
+            return self
```
vinceatbluelabs, line 318: Since we'll be calling this on stuff where it may not be necessary.

tests/integration/records/expected_column_types.py
```diff
@@ -134,8 +140,8 @@
     ],
     ('bigquery', 'redshift'): [
         'BIGINT', 'VARCHAR(256)', 'VARCHAR(256)', 'VARCHAR(256)', 'VARCHAR(256)',
-        'VARCHAR(256)', 'VARCHAR(256)', 'DATE', 'VARCHAR(8)',
-        'TIMESTAMP WITHOUT TIME ZONE', 'TIMESTAMP WITH TIME ZONE'
+        'VARCHAR(256)', 'VARCHAR(256)', 'VARCHAR(256)', 'VARCHAR(8)', 'VARCHAR(256)',
+        'VARCHAR(256)'
```
vinceatbluelabs, line 144: Downside: No date/time types.

Upside: Can handle larger tables, as we're not working through SELECT and Pandas!

records_mover/db/redshift/redshift_db_driver.py
```diff
@@ -154,3 +156,36 @@ def loader_from_fileobj(self) -> None:
 
     def unloader(self) -> Optional[Unloader]:
         return self._redshift_unloader
+
+    def tweak_records_schema_for_load(self,
+                                      records_schema: RecordsSchema,
+                                      records_format: BaseRecordsFormat) -> RecordsSchema:
+        if isinstance(records_format, AvroRecordsFormat):
```
cwegrzyn, line 163: Any thoughts on whether it's appropriate to warn the user here that the types are being downcast? Not sure if that's come up before (or what the exact right mechanism/location for the logic to be), but it might be a helpful hint that "oh hey, your date column is confusingly going to be an integer of days from the epoch". (Might also keep us on our toes for finding ways to improve it in the future!)

records_mover/records/targets/data_url.py
```diff
@@ -13,13 +15,22 @@
     from ..sources.dataframes import DataframesRecordsSource
 
 
+def sniff_records_format_from_url(url: str) -> BaseRecordsFormat:
+    if url.endswith('.parquet'):
```
cwegrzyn, line 19: Would it ever be the case that we have .parquet.gz or .avro.gz files? I think compression is part of the file format in those formats so the answer is "no", but just verifying.

cwegrzyn (approved):
A few questions and an idea about warning users. But generally this looks great! (And once again shows how easy it is to expand the functionality in here in nontrivial ways!) :shipit:

PR #140, opened 2020-12-04
OBE: More example-based testing of date/time types

vinceatbluelabs:
This extends the pattern in #144 to test more code that processes hints.

PR #144, opened 2020-12-09
Loosen validation and mypy type requirements for date/time types, start increased testing

vinceatbluelabs:
The Records Spec uses hints to describe how dates, times, datetime, and datetimetz types are formatted in CSVs.

It does this by describing a set of allowed values. The format of the values were originally cribbed from Redshift's COPY statement.

The current set of actually allowed values for those types (dateformat/timeonlyformat/datetimeformattz/datetimeformat) are fairly arbitrary, incomplete, inconsistent with each other, and don't hold up as terribly consistent under scrutiny. They've also been a source of bugs over time.

To make those complete even insomuch as the patterns they currently try to support would involve the start of some combinatorial explosion:

* Hour can be specified as HH, HH24 or HH12 (note also that HH and HH24 mean the same thing, but we documented it both ways in at least one case.)
* SS can either be specified or not.
* OF can be specified for timezone offset or not
* Dates can be in MDY, DMY or YMD format
* Years can be in YYYY or YY format
* Dates can be segmented by - or /
* ...

I think this does not bode well for maintaining an allowlist of formats - instead, I think we're going to need to (like Redshift) support a general formatting language in the spec, and do translations as needed to the formatting languages of the different databases/tools that Records Mover integrates with.

In the face of that, I'd like to expand the allowed values in a future PR. The motivation there is to ensure we allow likely DMY-flavored formats, to more generally support more things users may need in the future.

This PR is step one of that. It adds one additional test as a beginning of comprehensive testing of the existing supported values, and makes it easy to add more testable values in the future.

To deal with the combinatorial explosion problem for the time being, I've loosened the internal constraints by changing the Literal types to strings, and fixed up the various things based on what the testing found.

This PR does not yet loosen up the spec in any way.

The rest of the tests can be found in #140, which is probably too big to review right now--this is the first PR chipped off of that boulder.


records_mover/db/bigquery/load_job_config_options.py
```diff
@@ -324,13 +324,14 @@ def add_load_job_csv_config(unhandled_hints: Set[str],
...
     if hints.datetimeformattz in ['YYYY-MM-DD HH:MI:SSOF',
+                                  'YYYY-MM-DD HH24:MI:SSOF',
                                   'YYYY-MM-DD HH:MI:SS']:
         pass
     else:
         cant_handle_hint(fail_if_cant_handle_hint, 'datetimeformattz', hints)
     quiet_remove(unhandled_hints, 'datetimeformattz')
 
-    if hints.timeonlyformat == 'HH24:MI:SS':
+    if hints.timeonlyformat in ['HH24:MI:SS', 'HH:MI:SS']:
```
vinceatbluelabs, line 334: This fixes issues found during testing!

tests/component/records/datetime_cases.py
```diff
@@ -0,0 +1,60 @@
...
+    "YYYY-MM-DD HH24:MI:SSOF",
+    "MM/DD/YY HH24:MI",
+]
+
+DATETIME_CASES = [
+    "YYYY-MM-DD HH24:MI:SS",
+    "YYYY-MM-DD HH:MI:SS",
+    "YYYY-MM-DD HH12:MI AM",
+    "MM/DD/YY HH24:MI",
+]
```
vinceatbluelabs, line 60: Not all of this file is probably used in this PR - it's the full set of things needed to define the test cases for #140

tests/component/db/bigquery/test_load_config_options_datetime.py
```diff
@@ -0,0 +1,132 @@
...
+            load_plan = RecordsLoadPlan(processing_instructions=processing_instructions,
+                                        records_format=records_format)
+            unhandled_hints = set(records_format.hints.keys())
+            try:
+                load_job_config(unhandled_hints, load_plan)
+            except NotImplementedError:
+                if should_raise[dateformat]:
+                    pass
+                else:
+                    raise
```
vinceatbluelabs, line 43: The general intent with these tests is that as we add new things to the below datetime_cases.py module, tests will blow up and we'll need to go back and hand-add expectations of what each driver should produce for each input.

records_mover/db/bigquery/load_job_config_options.py
vinceatbluelabs: Specifically, the Redshift spec for date/time formats defines HH as equivalent to HH24 - so there's no reason now to handle it similarly to the HH24 cases.  These were reproduced with test cases defined in datetime_cases.py in tests/component below.

records_mover/db/bigquery/load_job_config_options.py
```diff
@@ -324,13 +324,14 @@ def add_load_job_csv_config(unhandled_hints: Set[str],
     quiet_remove(unhandled_hints, 'datetimeformat')
 
     if hints.datetimeformattz in ['YYYY-MM-DD HH:MI:SSOF',
+                                  'YYYY-MM-DD HH24:MI:SSOF',
                                   'YYYY-MM-DD HH:MI:SS']:
```
vinceatbluelabs, line 328: `'YYYY-MM-DD HH24:MI:SS` would be a good additional item - I'll write this down and address it in #140 if not before, probably by adding it as an explicit test case in my list of test cases below.

vinceatbluelabs:
Looks like I have a test failure, which is annoying.  I’ll open up another PR related to that and probably merge master back into this branch afterwards.  If you could review the PR in the meantime ignoring the test failure for the moment, I’ll hold off merging until that fixes the passing tests.

crvena-sonja (approved):
The tests are a bit of a logical brain-twister here but I like the consistency and the comments on datatypes definitely help!

Nice cleanup here and helpful context, as always.

PR #145, opened 2020-12-09
Drop overly-complected Christmas tree tests

vinceatbluelabs:
To increase test coverage around part of the code which turn hints into config for various tools, I had used a test pattern I now regret.  This PR removes tests which use that pattern.

image


tests/component/db/vertica/test_vertica_import_options.py
```diff
@@ -87,34 +86,6 @@ def test_vertica_format_permissive(self):
...
-            'record_terminator': '\x02',
-            'rejectmax': 1,
-            'skip': 1,
-            'trailing_nullcols': False,
-        }
-        self.assertDictEqual(options, expected_options)
-        self.assertListEqual(mock_warning.mock_calls,
-                             [call("Ignoring hint compression = 'LZO'"),
-                              call("Ignoring hint quoting = 'nonnumeric'")])
-        self.assertEqual(unhandled_hints, set())
```
vinceatbluelabs, line 116: These tests try to validate things by a growing collection of eclectic hints which try to test a number of different things simultaneously.

Cool, I guess, but the downside is that debugging them when they fail is really hard and creates cognitive load.

Instead I'd like to test hints more individually - the tests in #144 and #140 are better examples of how I'd like to write these tests going forward.

I've been wanting to kill these for a while, but the forcing function was that I tried to debug these after making clearly safe changes and it gave me too much of a head-ache trying to detangle what was going on.

crvena-sonja (approved):
We're throwing the tree out rather early, but thank you for taking pity on those of us who would have tried to debug this later.... 😬 

:shipit:

metrics/coverage_high_water_mark
```diff
@@ -1 +1 @@
-93.6700
+93.1600
```
vinceatbluelabs, line 1: This is the downside.  I'll make up for some of that in #140, for sure, but won't have time to get us back to this number.

crvena-sonja: I do think that the complexity-reduction here is an overall benefit.

PR #150, opened 2020-12-09
Ensure close() is called on result of pandas.read_sql(chunksize=...)

vinceatbluelabs:
While writing new integration tests in #140, I started hitting frozen tests and 'too many open connection' errors from Redshift.  The patch inside this PR fixed it - it appears that we were not closing the SQL connection opened while reading chunked dataframes, which resulted in simultaneous connections piling up and table locks in the integration tests.


cwegrzyn (approved):
Ah, didn't know about generator `close`. But, interesting! :shipit:

PR #151, opened 2020-12-10
Improve date/time hint handling for MySQL based on better testing

vinceatbluelabs:
Based on the new testing approach in #144, this PR adds component tests for date/time types while loading CSV files in MySQL.

The expectations for those test are based on the results of the integration tests in #140, which shows our MySQL driver is way too optimistic about what date/time hints it can import!


cwegrzyn (approved):
:shipit:

PR #152, opened 2020-12-10
MySQL time column to dataframe fix for Pandas 1.x, increase MySQL integration testing

vinceatbluelabs:
MySQL doesn't really have a Time column type, it has a time delta type.  People however use that type for both things.

In order to format CSVs after pulling data from MySQL into a dataframe, we have some code which formats timedelta columns as time strings.  To do that, we were relying on some shady __str__() behavior in Pandas/NumPy timedelta types. 

It looks like as of Pandas 1.x, for the time midnight (`00:00:00`), this breaks, as str(timedelta) gives us `0 days` instead of `0 days 00:00:00`.

This PR changes over to a less shady method of generating those times.


.circleci/config.yml
```diff
@@ -465,6 +469,11 @@ workflows:
           name: mysql-itest
           extras: '[mysql,itest]'
           python_version: "3.6"
+          # Using Pandas reproduced a bug that happened when we were
+          # relying on Pandas:
+          #
+          # See https://github.com/bluelabsio/records-mover/pull/152
+          pandas_version: ">1"
```
vinceatbluelabs, line 476: Reproduced here: https://app.circleci.com/pipelines/github/bluelabsio/records-mover/1812/workflows/c7b38b71-8ef9-4f6c-b096-10fe82d0c2ca/jobs/15504

tests/integration/records/single_db/test_records_load.py
```diff
@@ -34,17 +34,21 @@ def load_and_verify(self, format_type, variant, hints={}, broken=False, sourcefn
...
-        self.load_and_verify('delimited', 'bigquery', {'header-row': True})
+        self.load_and_verify('delimited', 'bigquery', {'compression': None, 'header-row': True})
 
     def test_load_bluelabs_format(self):
-        self.load_and_verify('delimited', 'bluelabs')
+        self.load_and_verify('delimited', 'bluelabs', {'compression': None})
 
     def test_load_bluelabs_format_with_header_row(self):
-        self.load_and_verify('delimited', 'bluelabs', {'header-row': True})
+        self.load_and_verify('delimited', 'bluelabs', {'compression': None, 'header-row': True})
```
vinceatbluelabs, line 51: This is unrelated to the above concern - but we weren't really testing MySQL unloads, as our test cases used compressed files.  Those aren't supported by MySQL

tests/unit/records/schema/field/test_field.py
```diff
@@ -242,3 +242,19 @@ def test_cast_series_type_time_timedelta_entries(self):
...
+        mock_representations = Mock(name='representations')
+        field = RecordsSchemaField(name=mock_name,
+                                   field_type=mock_field_type,
+                                   constraints=mock_constraints,
+                                   statistics=mock_statistics,
+                                   representations=mock_representations)
+        data = np.array([pd.Timedelta(hours=0, minutes=0, seconds=0)])
+        series = pd.Series(data)
+        new_series = field.cast_series_type(series)
+        self.assertEqual(new_series[0], '00:00:00')
```
vinceatbluelabs, line 260: This test reproduced the problem:

https://app.circleci.com/pipelines/github/bluelabsio/records-mover/1813/workflows/888be2a6-5694-4ddb-8436-7f6a2c386948/jobs/15507

records_mover/records/schema/field/__init__.py
```diff
@@ -185,8 +185,13 @@ def cast_series_type(self, series: 'Series') -> 'Series':
...
                 #
                 if type(series[0]) == pd.Timedelta:
+                    # https://stackoverflow.com/questions/34501930/how-to-convert-timedelta-to-time-of-day-in-pandas
+
                     # Convert from "0 days 12:34:56.000000000" to "12:34:56"
-                    return series.astype(str).str.split().str[-1].str.split('.').str[0]
+                    def components_to_time_str(df: pd.DataFrame) -> str:
+                        return f"{df['hours']:02d}:{df['minutes']:02d}:{df['seconds']:02d}"
+                    out = series.dt.components.apply(axis=1, func=components_to_time_str)
+                    return out
```
vinceatbluelabs, line 194: I don't know how much more or less efficient this will be.  It's probably not hugely important until we start dealing with very large things exported from MySQL - and given we don't currently have a bulk unloader, adding one is probably a higher priority than trying to optimize this.  🤷

.circleci/config.yml
cwegrzyn: Is there any sense in running these tests with old pandas as well? I mean, hopefully people aren't really using old pandas much anymore (should that maybe become the default pandas version unless overridden?) I guess that's a lot of testing to do for a relatively small segment of functionality

records_mover/records/schema/field/__init__.py
cwegrzyn: Word-- I think my expectation is that before we worry about this time transformation in pandas being a bottleneck for big datasets, we probably need to worry about pandas itself being the bottleneck.

tests/integration/records/single_db/test_records_load.py
cwegrzyn: Should that read "we _aren't_ really testing..."? If so, I will drop an issue in for us to add some testing :)

cwegrzyn (approved):
Nice! A few small questions, but nothing to hold up the PR. :shipit:

.circleci/config.yml
vinceatbluelabs: Honestly I'm only halfway convinced it makes sense to keep this in the tests as a special case - this one line of code is currently the issue, there's nothing else special about MySQL and Pandas, and now we have a unit test for this issue, which is lower on the test pyramid (and we run unit tests on both Pandas <1 and Pandas >1).

PR #153, opened 2020-12-10
Fix Pandas date formatting, add date/time integration tests

vinceatbluelabs:
While we now have component tests for date/time types, they are only as good as the manual testing which was done to prove the behavior.  In a couple of cases, that hasn't been enough.

This PR adds automated integration tests for all of the date/time types for each database type we have, for both loading and unloading.

It also fixes an issue , where if an export ends up being done via Pandas, date format requests are ignored.


tests/integration/records/single_db/test_records_unload_datetime.py
```diff
@@ -0,0 +1,363 @@
...
+                                 f"from dateformat {dateformat} and addl_hints {addl_hints}")
+            except ModuleNotFoundError as e:
+                if 'pandas' in str(e) and expect_pandas_failure:
+                    # as expected
+                    continue
+                else:
+                    raise
+            self.assertFalse(expect_pandas_failure)
+
+    def test_unload_datetime(self) -> None:
```
bluelabsbutler, line 92: C901 'RecordsUnloadDatetimeIntegrationTest.test_unload_datetime' is too complex (17)

tests/integration/records/single_db/test_records_unload_datetime.py
```diff
@@ -0,0 +1,363 @@
...
+            except NotImplementedError as e:
+                if expect_am_failure and 'AM' in str(e):
+                    # as expected
+                    continue
+                else:
+                    raise
+            self.assertFalse(expect_pandas_failure)
+            self.assertFalse(expect_am_failure)
+
+    def test_unload_datetimetz(self) -> None:
```
bluelabsbutler, line 185: C901 'RecordsUnloadDatetimeIntegrationTest.test_unload_datetimetz' is too complex (17)

tests/integration/records/single_db/test_records_unload_datetime.py
```diff
@@ -0,0 +1,363 @@
...
+            except NotImplementedError as e:
+                if expect_am_failure and 'AM' in str(e):
+                    # as expected
+                    continue
+                else:
+                    raise
+            self.assertFalse(expect_pandas_failure)
+            self.assertFalse(expect_am_failure)
+
+    def test_unload_timeonly(self) -> None:
```
bluelabsbutler, line 289: C901 'RecordsUnloadDatetimeIntegrationTest.test_unload_timeonly' is too complex (17)

records_mover/records/pandas/prep_for_csv.py
```diff
@@ -35,10 +34,13 @@ def _convert_series_or_index(series_or_index: T,
...
-            if isinstance(series_or_index, pd.Series):
-                return series_or_index.dt.strftime(pandas_date_format)
+            if isinstance(series_or_index[0], pd.Timestamp):
+                if isinstance(series_or_index, pd.Series):
+                    return series_or_index.dt.strftime(pandas_date_format)
+                else:
+                    return series_or_index.strftime(pandas_date_format)
             else:
-                return series_or_index.strftime(pandas_date_format)
+                return series_or_index.apply(pd.Timestamp).dt.strftime(pandas_date_format)
```
vinceatbluelabs, line 43: This bug was reproduced by the integration tests:

https://app.circleci.com/pipelines/github/bluelabsio/records-mover/1823/workflows/ef8242cd-55c3-4442-b72b-37bea71c89eb/jobs/15583


Which I wrote a component test for:

https://app.circleci.com/pipelines/github/bluelabsio/records-mover/1825/workflows/9249f424-617c-409e-a3de-e57d62ea811e/jobs/15587

That then revealed we had workaround expectation files that can be fixed to be more correct!

https://app.circleci.com/pipelines/github/bluelabsio/records-mover/1828/workflows/54774c27-e141-40ea-bb32-9d7123ff766d/jobs/15594

tests/integration/records/single_db/test_records_unload_datetime.py
vinceatbluelabs: I haven't tried much further to refactor these to reduce complexity, and quality is failing as a result.  Any suggestions?  I'm inclined to bump the numbers and merge and either declare this to be the best trade-off for now or make this a problem for the future.

tests/integration/records/single_db/base_records_test.py
```diff
@@ -26,10 +29,7 @@
 logger = logging.getLogger(__name__)
 
 
-stdout_handler = logging.StreamHandler(stream=sys.stdout)
-formatter = logging.Formatter('%(asctime)s - %(message)s', '%H:%M:%S')
-stdout_handler.setFormatter(formatter)
-logging.getLogger().addHandler(stdout_handler)
+set_stream_logging()
```
vinceatbluelabs, line 32: While debugging having better logging was helpful!

tests/integration/records/single_db/test_records_unload_datetime.py
cwegrzyn: Not immediately, so I think bumping is fine.

The one thing that came to mind-- but I don't know the right abstraction off the top of my head-- is that fundamentally stuff like "use_pandas" comes down to some facts about support that the database and/or records-mover have for certain functionality or behaviors. Perhaps there are some decisions that can be pushed into standard interfaces in drivers instead-- that could reduce some of the branching.

The most straightforward idea is something like `will_use_pandas_for_unload`. That seems a little bit like exposing implementation details (and making it the problem of the end user, such as this test), rather than exposing the actual constraints. But maybe it's still better than baking the logic into the tests?

cwegrzyn (approved):
Another nice find. And also more yikes from the world of dates and times. Had one half thought about the test complexity-- but I think it's also okay to bump now, unless you really want to rack your brain for the right abstraction on your last day! :shipit:

PR #158, opened 2021-01-20
Loosen tenacity dependency

cwegrzyn:
Loosen the version of tenacity required to improve compatibility with other deps (such as airflow) especially under new pip resolver. I did a comparison of the tenacity code between 4.12 and 6.0, and didn't see anything that we relied on.


setup.py
```diff
@@ -134,7 +134,9 @@ def initialize_options(self) -> None:
 )
 
 airflow_dependencies = [
-    'apache-airflow>=1.10,<2'
+    # Minimum version here is needed to avoid syntax error in setup.py
```
bluelabsbutler, line 137: This file, one of the 3 largest in the project, increased in size to 379 lines.  The total size of those files is now 3421 lines (target: 1147).  Is this file complex enough to refactor?

bluelabsbutler, line 137: This file, one of the 3 largest in the project, increased in size to 379 lines.  The total size of those files is now 3421 lines (target: 1149).  Is this file complex enough to refactor?

PR #169, opened 2021-10-07
Update mypy to latest

cwegrzyn:


setup.py
```diff
@@ -198,8 +198,7 @@ def initialize_options(self) -> None:
 ] + db_dependencies
 
 redshift_dependencies_base = [
-    # sqlalchemy-redshift 0.7.7 introduced support for Parquet
-    # in UNLOAD
+    # sqlalchemy-redshift 0.7.7 introduced support for Parquet in UNLOAD
```
bluelabsbutler, line 201: This file, one of the 3 largest in the project, increased in size to 365 lines.  The total size of those files is now 3407 lines (target: 1135).  Is this file complex enough to refactor?

PR #170, opened 2022-03-03
Airflow 2.0 Compatibility

archetypalsxe:
Added a try/catch around the Airflow BaseHook import to add compatibility with Airflow 2.0


cwegrzyn (approved):
:shipit:

PR #178, opened 2022-12-12
Increase CircleCI Docker resource classes to Large

ryantimjohn:


ryantimjohn:
This does not seem to lower build times. Going to put this one to bed.

PR #180, opened 2022-12-14
Update all Python 3.6 tests to 3.9

ryantimjohn:


setup.py
```diff
@@ -126,6 +126,7 @@ def initialize_options(self) -> None:
 
 itest_dependencies = [
     'jsonschema',  # needed for directory_validator.py
+    'pytz',
```
bluelabsbutler, line 129: This file, one of the 3 largest in the project, increased in size to 366 lines.  The total size of those files is now 3408 lines (target: 1135).  Is this file complex enough to refactor?

setup.py
```diff
@@ -126,6 +126,8 @@ def initialize_options(self) -> None:
...
```
bluelabsbutler, line 129: This file, one of the 3 largest in the project, increased in size to 367 lines.  The total size of those files is now 3409 lines (target: 1136).  Is this file complex enough to refactor?

setup.py
```diff
@@ -126,6 +126,8 @@ def initialize_options(self) -> None:
...
+    'wheel', # needed to support legacy 'setup.py install'
```
bluelabsbutler, line 130: E261 at least two spaces before inline comment

setup.py
bluelabsbutler, line 129: This file, one of the 3 largest in the project, increased in size to 367 lines.  The total size of those files is now 3409 lines (target: 1137).  Is this file complex enough to refactor?

records_mover/records/targets/google_sheets.py
```diff
@@ -104,7 +104,7 @@ def _get_service(self) -> SheetsService:
     def as_json_serializable(self, cell: Any) -> Any:
         if isinstance(cell, np.generic):
             # MyPy complains that this method does not exist
-            native = np.asscalar(cell)  # type: ignore
+            native = cell.item()
```
bluelabsbutler, line 107: W291 trailing whitespace

.circleci/config.yml
```diff
@@ -568,7 +568,7 @@ workflows:
       - integration_test:
           name: redshift-s3-itest-old-pandas
           extras: '[redshift-binary,itest]'
-          python_version: "3.9"
+          python_version: "3.8"
```
Brunope, line 571: why is this being changed to 3.8?

Brunope: Ah I'm guessing it's because this test specifically is targeting compatibility with an old version of pandas that maybe requires an older python version? the rest of these have been bumped to 3.9 as expected.

Brunope (approved):

mitchellj (approved):

.circleci/config.yml
ryantimjohn: you got it!

ryantimjohn: Tried to find a way to get around this by allowing it to build the old version of pandas in 3.9 but kept creating compounding errors

PR #181, opened 2022-12-15
RM-3 fix postgres integration tests

Brunope:
Opening PR to see if circleci works

PR #182, opened 2022-12-16
RM-17-fix-flake-8-violations

ryantimjohn:
Fix Flake8 violations to make future development easier


setup.py
```diff
@@ -126,6 +126,8 @@ def initialize_options(self) -> None:
 
 itest_dependencies = [
     'jsonschema',  # needed for directory_validator.py
+    'pytz',
```
bluelabsbutler, line 129: This file, one of the 3 largest in the project, increased in size to 367 lines.  The total size of those files is now 3409 lines (target: 1137).  Is this file complex enough to refactor?

records_mover/records/records_format.py
```diff
@@ -68,9 +68,9 @@ class DelimitedRecordsFormat(BaseRecordsFormat):
     in the constructor"""
 
     def __init__(self,
-                 variant: str='bluelabs',
-                 hints: PartialRecordsHints={},
-                 processing_instructions: ProcessingInstructions=ProcessingInstructions()) -> None:
+                 variant: str = 'bluelabs',
+                 hints: PartialRecordsHints = {},
+                 processing_instructions: ProcessingInstructions = ProcessingInstructions()) -> None:
```
bluelabsbutler, line 73: E501 line too long (101 > 100 characters)

tests/unit/records/targets/table/test_target.py
```diff
@@ -31,8 +31,7 @@ def setUp(self):
                                add_user_perms_for=self.mock_add_user_perms_for,
                                add_group_perms_for=self.mock_add_group_perms_for,
                                existing_table_handling=self.mock_existing_table_handling,
-                               drop_and_recreate_on_load_error=
-                               self.mock_drop_and_recreate_on_load_error)
+                               drop_and_recreate_on_load_error=self.mock_drop_and_recreate_on_load_error)
```
bluelabsbutler, line 34: E501 line too long (105 > 100 characters)

tests/unit/records/targets/table/test_target.py
```diff
@@ -32,7 +32,7 @@ def setUp(self):
                                add_group_perms_for=self.mock_add_group_perms_for,
                                existing_table_handling=self.mock_existing_table_handling,
                                drop_and_recreate_on_load_error=
-                               self.mock_drop_and_recreate_on_load_error)
+                                self.mock_drop_and_recreate_on_load_error)
```
bluelabsbutler, line 35: E127 continuation line over-indented for visual indent

ryantimjohn:
@Brunope @bcastro-bl I've merged the changes from the other two PRs #180 and #181 in here so that we can merge this one once those other two are in

ryantimjohn:
These are all the "low lift" flake violations. The other ones are of the "{foo} is too complex" sort which I feel like can be handled in another PR.

records_mover/db/driver.py
```diff
@@ -83,7 +83,7 @@ def set_grant_permissions_for_users(self, schema_name: str, table: str,
     def supports_time_type(self) -> bool:
         return True
 
-    def type_for_date_plus_time(self, has_tz: bool=False) -> sqlalchemy.sql.sqltypes.DateTime:
+    def type_for_date_plus_time(self, has_tz: bool = False) -> sqlalchemy.sql.sqltypes.DateTime:
```
Brunope, line 86: From what I've seen so far the original codebase more commonly uses the first option, with no extra space (my personal preference as well), but not a big deal.

records_mover/records/prep_and_load.py
```diff
@@ -15,7 +15,7 @@ def prep_and_load(tbl: TargetTableDetails,
                   load: Callable[[DBDriver], Optional[int]],
                   load_exception_type: Type[Exception],
                   reset_before_reload: Callable[[], None] = lambda: None) -> MoveResult:
-    logger.info(f"Connecting to database...")
+    logger.info("Connecting to database...")
```
Brunope, line 18: prefer single quotes maybe?

records_mover/records/targets/google_sheets.py
```diff
@@ -104,7 +104,7 @@ def _get_service(self) -> SheetsService:
     def as_json_serializable(self, cell: Any) -> Any:
         if isinstance(cell, np.generic):
             # MyPy complains that this method does not exist
-            native = np.asscalar(cell)  # type: ignore
+            native = cell.item()
```
Brunope, line 107: can't add this comment at the exact line but it seems like there should be a way to use numpy.ndarray.tobytes or something similar to serialize this instead of the nested list comprehension calling this method down below.

metrics/flake8_high_water_mark
```diff
@@ -1 +1 @@
-166
+9
```
Brunope, line 1: I don't exactly understand how these are used but why is this such a large numerical change?

Brunope (approved):
I don't exactly understand the flake8 water mark change but other than that looks good to me.

metrics/flake8_high_water_mark
ryantimjohn: So these "high water marks" are the number of violations there were of a particular protocol. In this case, we had 166 Flake8 violations and brought them down to 9. I mostly did this through using `autopep8` then hand correcting some of the other errors. We still have 9 flake8 errors remaining... but those are from "too high complexity" which is more than I wanted to take on in a single PR.

records_mover/db/driver.py
ryantimjohn: Agreed stylistically but triggering flake8 so 🤷

records_mover/records/prep_and_load.py
ryantimjohn: I've never gotten a good call on which to use and don't care much but... preserving the formatting here and also this resource recommends that, generally, use doubles for sentences because you might add a contraction, i.e. `"You've got an error!"` and bork your single quotes 🤷 :
https://docs.ckan.org/en/ckan-2.1.5/python-coding-standards.html#use-single-quotes

records_mover/records/targets/google_sheets.py
ryantimjohn: definitely agree; this is actually just a change I pulled forward from the `RM-7-update-default-python-image-for-integration-tests-to-3-9-remove-build-tests-for-deprecated-python-version` branch where there were so many errors on transition that I was trying to keep changes as minimal as possible. 

*But,* if you're willing to dive deep and rewrite this call, just create a ticket for it and it'll be a good way to get to know the tool better! We'll fix the failing tests first but then the world is our oyster. =)

PR #188, opened 2022-12-21
Add github actions job to run unit tests for 3.8, 3.9

Brunope:
Currently I think this runs against any commit even if no PR is open on the branch. Can change that if we want.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,7 +1,7 @@
 name: Test
 
 run-name: Tests
-on: [push]
+on: [pull_request]
 
 jobs:
   unit-tests:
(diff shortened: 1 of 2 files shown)


Brunope:
Would like to merge this - in another branch I'm planning to add a CI test to validate the readthedocs build and I'd like to add it in github actions form

Brunope:
You're right I think the behavior should be slightly different. The pull-request event isn't quite what we want because it wouldn't run on commits to an open pull request which we seem to rely on a lot. If we want to match the behavior of our circleci config we want it run on push events still but restrict it only to branches with open pull requests. I can make that change

Brunope:
Never mind the pull_request event is exactly what we want - if you push a commit which changes the HEAD ref of a pull request, it counts as a pull_request event with the synchronize action. Perfect!

PR #190, opened 2023-01-03
add 3.11 test

ryantimjohn:


setup.py
```diff
@@ -120,16 +120,17 @@ def initialize_options(self) -> None:
     'grpcio<2.0dev,>=1.29.0',
 ]
 
-nose_dependencies = [
-    'nose'
+pytest_dependencies = [
```
bluelabsbutler, line 123: This file, one of the 3 largest in the project, increased in size to 368 lines.  The total size of those files is now 3410 lines (target: 1137).  Is this file complex enough to refactor?

PR #191, opened 2023-01-05
RM-33-migrate-from-nose-to-pytest

ryantimjohn:
Migrate from nose to pytest

Additionally upgrade from deprecated convenience images as outlined here:
https://bluelabsio.atlassian.net/jira/software/projects/RM/boards/3?selectedIssue=RM-6


setup.py
```diff
@@ -120,16 +120,17 @@ def initialize_options(self) -> None:
     'grpcio<2.0dev,>=1.29.0',
 ]
 
-nose_dependencies = [
-    'nose'
+pytest_dependencies = [
```
bluelabsbutler, line 123: This file, one of the 3 largest in the project, increased in size to 368 lines.  The total size of those files is now 3410 lines (target: 1137).  Is this file complex enough to refactor?

bluelabsbutler, line 123: This file, one of the 3 largest in the project, increased in size to 368 lines.  The total size of those files is now 3410 lines (target: 1138).  Is this file complex enough to refactor?

requirements.txt
```diff
@@ -6,7 +6,7 @@ flake8
 # https://github.com/pallets/markupsafe/issues/284
 markupsafe==2.0.1
 # https://github.com/pandas-dev/pandas/pull/45749
-pandas==1.1.5
+# pandas
```
Brunope, line 9: I don't think this is right, why are we removing pandas here?

.circleci/config.yml
```diff
@@ -669,3 +671,5 @@ workflows:
               only: /v\d+\.\d+\.\d+(-[\w]+)?/
             branches:
               ignore: /.*/
+
+# VS Code Extension Version: 1.5.1
```
Brunope, line 675: is this something you meant to commit?

setup.py
```diff
@@ -192,7 +193,7 @@ def initialize_options(self) -> None:
 ]
 
 pandas_dependencies = [
-    'pandas==1.1.5',
+    'pandas',
```
Brunope, line 196: This is another change that I'm not super comfortable with - why was it pinned at 1.1.5 in the first place, and are we really sure this will work with ANY pandas version? The circleci config currently explicitly tests against pandas 1.1.5 and 1.3.5 (?), I think at minimum we would want to expand that to maybe testing against 1.1.5 and whatever the latest version is.

Brunope (changes requested):
Not sure about the pandas thing, I'd prefer if we didn't change anything related to pandas versioning in this pull request in order to minimize possible breakage.

Brunope:
On second thought, I'm okay with removing the versioning constraint on pandas. But I do still think it belongs in requirements.txt, because it definitely is a development requirement.

.circleci/config.yml
ryantimjohn: nope!

requirements.txt
ryantimjohn: I think you're right about this; having an explicit version is a liability as we go to future versions of python but, until we're there happy to take this out.

setup.py
ryantimjohn: I think you're right about this; having an explicit version is a liability as we go to future versions of python but, until we're there happy to take this out.

ryantimjohn:
Agree with having pandas version explicit in the reqs; I started this off of a branch where I was working on python 3.11 compatibility and those changes aren't relevant to the goal of moving to pytest!

As for explicit pandas versioning, I think I know why things are that way. Around 1.2, pandas mistakenly added an infinite recursion problem that very rarely came up, except in records mover. =) So we had a bunch of people in the #help channel asking about it. Pandas 1.1.5 and 1.4 got rid of the error, so I think now we can just allow people to build off the most modern version. /But/, will leave that change for trying to upgrade to allowing Python 3.11.

ryantimjohn:
@Brunope made requested changes, let me know if anything else needed!

PR #192, opened 2023-01-06
Rm 34 upgrade syntax to support airflow 2 0

ryantimjohn:


setup.py
```diff
@@ -120,16 +120,17 @@ def initialize_options(self) -> None:
     'grpcio<2.0dev,>=1.29.0',
 ]
 
-nose_dependencies = [
-    'nose'
+pytest_dependencies = [
```
bluelabsbutler, line 123: This file, one of the 3 largest in the project, increased in size to 368 lines.  The total size of those files is now 3410 lines (target: 1138).  Is this file complex enough to refactor?

PR #195, opened 2023-01-15
RM-21-Add-tests-for-Python-3.10-build

ryantimjohn:


records_mover/records/schema/field/representation.py
```diff
@@ -56,7 +56,7 @@ def from_data(data: 'FieldRepresentationDict') -> Optional['RecordsSchemaFieldRe
     def from_series(series: 'pandas.Series') -> 'RecordsSchemaFieldRepresentation':
         import pandas
 
-        dtype_json_str = pandas.io.json.dumps(series.dtype)
+        dtype_json_str = pandas.io.json.dumps(series.dtype,default_handler=str)
```
bluelabsbutler, line 59: E231 missing whitespace after ','

requirements.txt
```diff
@@ -6,7 +6,7 @@ flake8
 # https://github.com/pallets/markupsafe/issues/284
 markupsafe==2.0.1
 # https://github.com/pandas-dev/pandas/pull/45749
-pandas==1.1.5
+pandas
```
Brunope, line 9: Can you give some context on why this changed? I'm assuming pandas 1.1.5 doesn't work with python3.10? Regardless we definitely need to still specify a minimum version of pandas (applies to setup.py too), but I don't know what that version is.

ryantimjohn: So there's no pre-compiled wheel for pandas 1.1.5 on python 3.10, so when trying to build it from source the `Install python deps in venv` step would spin on forever (don't have a good illustration here on circleci since I did this in the SSH but this is an illustration of it running for at least 4 minutes):
https://app.circleci.com/pipelines/github/bluelabsio/records-mover/2234/workflows/362482c7-4ce4-40d1-8259-28839833798a/jobs/18590

Therefore, I decided to install this without a version number. I think you're right that it makes more sense to make this >= 1.1.5 as 1.1.5's minimum version is lower than the lowest currently supported python version:
https://pypi.org/project/pandas/1.1.5/#files
https://devguide.python.org/versions/

tests/integration/records/records_schema_v1_schema.json
```diff
@@ -1,294 +1,58 @@
 {
```
ryantimjohn, line 1: This is a weird way to validate whether a json is correctly formatted but is the way that we're currently doing it.

Brunope (approved):

PR #200, opened 2023-01-22
RM-22-Add-CircleCI-tests-for-Python-3.11-build

ryantimjohn:


ryantimjohn:
@Brunope ready for review, but should merge #195 first

tests/integration/records/records_schema_v1_schema.json
```diff
@@ -1,294 +1,58 @@
 {
```
Brunope, line 1: I don't understand what's going on in this file but I'll trust you that it's all good

tests/unit/records/schema/test_pandas.py
```diff
@@ -83,7 +83,7 @@ def test_pandas_numeric_types_and_constraints(self):
             ('float32', np.float32),
             ('float64', np.float64),
             # 'float96', np.float96), # not supported by numpy on macOS on amd64, apparantly
-            ('float128', np.float128),
+            # ('float128', np.float128),
```
Brunope, line 86: Just delete it if it's not gonna be used, but are there some platforms that do use this functionality still?

Brunope (approved):
Same concern as RM-21 regarding pandas versioning but otherwise looks good

PR #204, opened 2023-02-02
RM-13-fix-vertica-no-s-3-integration-test

ryantimjohn:


ryantimjohn:
@Brunope well, uh... this just works now. =)
Assume this is because of the changes we made for other DBs fixing whatever was wrong with Vertica.
I pulled in the changes from #203 just to make the tests run quicker, we can merge that one first.

Brunope (approved):

Issue #197, opened 2023-01-19
Upgrade Tenacity Requirement


archetypalsxe:
Currently the tenacity version is set to <7, >=4.12.0, however Airflow 2.2.2 requires version 8.0.1. Could we test with tenacity version 8.0.1 and allow that version?

Brunope:
Yeah I'll make a ticket for it

archetypalsxe:
Thank you!

PR #207, opened 2023-02-06
RM-36 bump tenacity version

ryantimjohn:


ryantimjohn:
@Brunope, seems like this just worked.

Brunope:
Have you tested this specifically with airflow >= 2.2.2, since that was the reason for this upgrade? The setup.py still says ">=2.0". If you have, looks good!

Brunope (approved):

ryantimjohn:
Screenshot 2023-02-08 at 10 57 07 AM
Yep! Will chat in our call today just to make sure I understand @archetypalsxe 's original ticket!

ryantimjohn:
Related to #197

ryantimjohn:
Okay... this is dumb. So, the coverage is different between different versions of python with this particular version of Airflow. 😠 

Otherwise this passes. 

I'm going to revert to >=2.0; if this ever comes up again, we'd have to update this class to be able to handle ranges:
https://github.com/bluelabsio/records-mover/blob/9167f38563a8c105ac9e9e755a8ac5b2219b00e8/setup.py#L39-L73

Or find a different way to handle ratcheting and all that which didn't rely on development in-house.

Brunope:
I'm pretty annoyed  by the ratchet coverage tests too, see this PR https://github.com/bluelabsio/records-mover/pull/209 where changing a version from 39.0.0 to 39.0.1 breaks the coverage test.

setup.py
```diff
@@ -8,6 +8,7 @@
 from distutils.cmd import Command
 import os.path
 import sys
+import warnings
```
ryantimjohn, line 11: This file, one of the 3 largest in the project, increased in size to 362 lines.  The total size of those files is now 3404 lines (target: 1132).  Is this file complex enough to refactor?

PR #208, opened 2023-02-06
RM-39 update to sqlalchemy 1.4

ryantimjohn:


tests/integration/records/single_db/test_records_numeric.py
```diff
@@ -67,8 +67,15 @@ def test_numeric_schema_fields_created(self) -> None:
             self.validate_records_schema(tempdir)
 
     def validate_table(self):
-        columns = self.engine.dialect.get_columns(self.engine, self.table_name,
-                                                  schema=self.schema_name)
+        if self.target_db_engine.driver == 'pymysql':
+            conn = self.target_db_engine.connect()
+            columns = self.target_db_engine.dialect.get_columns(conn,
+                                                                table_name,
```
ryantimjohn, line 73: F821 undefined name 'table_name'

tests/integration/records/single_db/test_records_numeric.py
```diff
@@ -67,8 +67,15 @@ def test_numeric_schema_fields_created(self) -> None:
...
+                                                                schema=schema_name)
```
ryantimjohn, line 74: F821 undefined name 'schema_name'

tests/integration/records/single_db/test_records_numeric.py
```diff
@@ -67,8 +67,15 @@ def test_numeric_schema_fields_created(self) -> None:
...
+        else:
+            columns = self.target_db_engine.dialect.get_columns(self.target_db_engine,
+                                                                table_name,
```
ryantimjohn, line 77: F821 undefined name 'table_name'

tests/integration/records/single_db/test_records_numeric.py
```diff
@@ -67,8 +67,15 @@ def test_numeric_schema_fields_created(self) -> None:
...
+                                                                schema=schema_name)
```
ryantimjohn, line 78: F821 undefined name 'schema_name'

tests/integration/records/single_db/test_records_numeric.py
```diff
@@ -67,8 +67,15 @@ def test_numeric_schema_fields_created(self) -> None:
             self.validate_records_schema(tempdir)
 
     def validate_table(self):
-        columns = self.engine.dialect.get_columns(self.engine, self.table_name,
-                                                  schema=self.schema_name)
+        if self.target_db_engine.driver == 'pymysql':
+            conn = self.target_db_engine.connect()
+            columns = self.engine.dialect.get_columns(conn,
+                                                                table_name,
```
ryantimjohn, line 73: E127 continuation line over-indented for visual indent

tests/integration/records/single_db/test_records_numeric.py
```diff
@@ -67,8 +67,15 @@ def test_numeric_schema_fields_created(self) -> None:
...
+                                                                schema=schema_name)
```
ryantimjohn, line 74: E127 continuation line over-indented for visual indent

tests/integration/records/single_db/test_records_numeric.py
```diff
@@ -67,8 +67,15 @@ def test_numeric_schema_fields_created(self) -> None:
...
+        else:
+            columns = self.engine.dialect.get_columns(self.target_db_engine,
+                                                                table_name,
```
ryantimjohn, line 77: E127 continuation line over-indented for visual indent

tests/integration/records/single_db/test_records_numeric.py
```diff
@@ -67,8 +67,15 @@ def test_numeric_schema_fields_created(self) -> None:
...
+                                                                schema=schema_name)
```
ryantimjohn, line 78: E127 continuation line over-indented for visual indent

tests/unit/records/sources/test_table.py
```diff
@@ -54,6 +54,38 @@ def test_to_dataframes_source(self,
                                    records_schema=mock_records_schema)
             mock_chunks.close.assert_not_called()
         mock_chunks.close.assert_called()
+    def test_to_dataframes_source_mysql(self,
```
ryantimjohn, line 57: E301 expected 1 blank line, found 0

tests/unit/records/sources/test_table.py
```diff
@@ -54,6 +54,43 @@ def test_to_dataframes_source(self,
                                    records_schema=mock_records_schema)
             mock_chunks.close.assert_not_called()
         mock_chunks.close.assert_called()
+
```
ryantimjohn, line 57: W293 blank line contains whitespace

records_mover/records/sources/table.py
```diff
@@ -77,9 +77,15 @@ def to_dataframes_source(self,
         db = self.driver.db
         records_schema = self.pull_records_schema()
 
-        columns = db.dialect.get_columns(db,
-                                         self.table_name,
-                                         schema=self.schema_name)
+        if isinstance(db, Engine) db.driver == 'pymysql':
```
ryantimjohn, line 80: E999 SyntaxError

ryantimjohn:
@Brunope I will clean up commit history a lot here, feel free to review now or once that's done!

tests/integration/records/table_validator.py
```diff
@@ -62,9 +62,15 @@ def validate(self,
...
-                                                            schema=schema_name)
+        if self.target_db_engine.driver == 'pymysql':
+            conn = self.target_db_engine.connect()
+            columns = self.target_db_engine.dialect.get_columns(conn,
+                                                                table_name,
+                                                                schema=schema_name)
+        else:
+            columns = self.target_db_engine.dialect.get_columns(self.target_db_engine,
+                                                                table_name,
+                                                                schema=schema_name)
```
Brunope, line 73: This is so weird to me, does the get_columns method take different types of objects depending on what the driver is????

Brunope: From the sqlalchemy documentation, get_columns says it always takes a connection as the first arg.
https://docs.sqlalchemy.org/en/14/core/internals.html#sqlalchemy.engine.Dialect.get_columns.

I'm sure you've already tried this, but we can't just pass the `conn` object to both?

Is this records-mover's fault for storing inconsistent types in its wrappers around sqlalchemy?

Brunope:
```
class DBDriver(metaclass=ABCMeta):
    def __init__(self,
                 db: Union[sqlalchemy.engine.Engine,
                           sqlalchemy.engine.Connection], **kwargs) -> None:
```
well this is weird

Brunope: Alright so I think the reason for this is that in some cases we want to instantiate a db_driver in the context of a transaction, as in this example from records/mover/records/prep_and_load.py
```python
    with tbl.db_engine.begin() as db:
        # This second transaction ensures the table has been created
        # before non-transactional statements like Redshift's COPY
        # take place.  Otherwise you'll get an error like:
        #
        #  Cannot COPY into nonexistent table
        driver = tbl.db_driver(db)
        try:
            import_count = load(driver)
        except load_exception_type:
            if not tbl.drop_and_recreate_on_load_error:
                raise
            reset_before_reload()
            with tbl.db_engine.begin() as db:
                driver = tbl.db_driver(db)
                prep.prep(schema_sql=schema_sql,
                          driver=driver,
                          existing_table_handling=ExistingTableHandling.DROP_AND_RECREATE)
                import_count = load(driver)
        return MoveResult(move_count=import_count, output_urls=None)
```
`begin()` starts a transaction and returns a connection object - naming it a "db" is pretty confusing.

Anyway, other times we don't care about transactions and don't create the connection object beforehand. I'm kind of confused about this, but sqlalchemy had this feature "connection-less execution" which basically let you execute stuff on an `engine` directly without needing a `connection` first, but they've been phasing it out: https://docs.sqlalchemy.org/en/14/core/connections.html#connectionless-execution-implicit-execution. Maybe this is why the code used to work.

I think the proper solution is to split the union from `DBDriver` into an optional `connection` argument and an `Engine`. If no `connection` is given it can create one from the `Engine`. Then, when a caller wants to do something like `get_columns`, they always use the `connection` attribute of `DBDriver`, not the `Engine`. This solves a bunch of confusion caused by examples such as above, where `target_db_engine` isn't actually an engine, it's a connection.

Should we track this work separately or do it as part of this ticket?

If we choose to put this work off until later, in the short term I think we should change the above code that I commented on (as well as the other similar occurrences) to read something like
```python
if isinstance(self.target.db_engine, sqlalchemy.engine.Connection):
    connection = self.target_db_engine
else:
    connection = self.target._db_engine.connect()
get_columns(connection)
```

ryantimjohn:
@Brunope, your investigations on all of this match up with my own. Only thing that I'll add is that, in the most recent versions of SqlAlchemy, the `.get_columns()` method for the MySQL dialect base moved to calls that only work against the Connection class whereas other dialects get columns using calls that work against the Engine class.

I think you're approach is correct in that it future proofs us for when other dialects use calls that only work against Connection. I was limiting this to just MySQL to limit side-effects on other SQL dialects but you're right that there _shouldn't_ be any.

tests/integration/records/single_db/test_records_numeric.py
```diff
@@ -67,8 +68,17 @@ def test_numeric_schema_fields_created(self) -> None:
...
+            connection = self.engine.connect()
+            columns = self.engine.dialect.get_columns(connection,
+                                                      self.table_name,
+                                                      schema=self.schema_name)
+            connection.close()
+        else:
+            columns = self.engine.dialect.get_columns(self.engine,
+                                                      self.table_name,
+                                                      schema=self.schema_name)
+
```
ryantimjohn, line 81: W293 blank line contains whitespace

tests/integration/records/table_validator.py
```diff
@@ -62,9 +62,17 @@ def validate(self,
...
+            connection = self.target_db_engine.connect()
+            columns = self.target_db_engine.dialect.get_columns(connection,
+                                                                table_name,
+                                                                schema=schema_name)
+            connection.close()
+        else:
+            columns = self.target_db_engine.dialect.get_columns(self.target_db_engine,
+                                                                table_name,
+                                                                schema=schema_name)
+
```
ryantimjohn, line 75: W293 blank line contains whitespace

tests/integration/records/table_validator.py
ryantimjohn: Implemented this approach.

Brunope (approved):

PR #210, opened 2023-02-13
Ignore flake8 complexity warning for pandas/read_csv_options

Brunope:


Brunope:
After looking at this function, I think it's best to leave as is. 

1. It's already close to being below the threshold for complexity
2. It has a ton of comments explaining why it is and works the way it does
3. It really does seem to be the logical place to store the reading of these options. In my opinion it would be more confusing if this functionality was spread out into different places.

ryantimjohn:
This all sounds right @Brunope!

ryantimjohn (approved):

PR #223, opened 2023-03-03
RM-49 attempt a type to func map

ryantimjohn:


ryantimjohn:
Alright @Brunope, this is passing (except in python 3.7 where lambda assignment i.e. `:=` assignment isn't allowed, doh)
https://peps.python.org/pep-0572/
But, as we mentioned in our call, this is actually less legible than the more complex call.

Therefore, I'm going to try to make this a lot simpler by having fully scoped functions for the complex lambdas, and just return the resulting datatypes for less complex functions, and see if that works + looks more palatable.

Brunope:
Sounds good!

ryantimjohn:
Okay @Brunope, this is ready for review!

I think at this point, both legibility and complexity have been improved.

records_mover/records/schema/field/sqlalchemy.py
```diff
@@ -124,75 +124,73 @@ def field_from_sqlalchemy_column(column: Column,
...
-            # HH:MM AM; should be same in bytes as chars at least in UTF-8
-            return sqlalchemy.sql.sqltypes.VARCHAR(8)
-    else:
-        raise NotImplementedError("Teach me how to handle records schema "
-                                  f"type {field.field_type}")
+    field_to_func_map = {
+        'integer': get_integer_type,
+        'decimal': get_decimal_type,
+        'boolean': cast(Callable[['RecordsSchemaField', 'DBDriver'], sqlalchemy.types.TypeEngine],
+                        lambda field, driver: sqlalchemy.sql.sqltypes.BOOLEAN()),
```
Brunope, line 171: is there a way to `cast()` the entries in the map at the end all at once instead of each time individually?

Brunope (approved):
This is a good improvement! I would personally prefer all the lambdas to be replaced by fully qualified functions (lots of small functions is a good thing and easy to trace), but I leave it up to your discretion.

ryantimjohn:
Oh the downsides to a merge queue =)

PR #224, opened 2023-03-08
RM-49 Reduce field_to_sqlalchemy_type complexity

ryantimjohn:


ryantimjohn:
@Brunope , I pulled out the typecasting from the function map. 

I think I do have a preference for keeping the one/two-liners as lambdas rather than having a ton ton of fully-scoped functions in the code. Seems like it'd get crowded.

records_mover/records/schema/field/sqlalchemy.py
```diff
@@ -167,30 +167,23 @@ def field_to_sqlalchemy_type(field: 'RecordsSchemaField',
...
-                     else sqlalchemy.sql.sqltypes.VARCHAR(8)),
-        'timetz': cast(Callable[['RecordsSchemaField', 'DBDriver'], sqlalchemy.types.TypeEngine],
-                       lambda field, driver: sqlalchemy.sql.sqltypes.TIME(timezone=True)
-                       if driver.supports_time_type()
-                       else sqlalchemy.sql.sqltypes.VARCHAR(8))
+        'date': lambda field, driver: sqlalchemy.sql.sqltypes.DATE(),
+        'datetime': lambda field, driver: driver.type_for_date_plus_time(has_tz=False),
+        'datetimetz': lambda field, driver: driver.type_for_date_plus_time(has_tz=True),
+        'time': lambda field, driver: sqlalchemy.sql.sqltypes.TIME(timezone=False)
+        if driver.supports_time_type() else sqlalchemy.sql.sqltypes.VARCHAR(8),
```
Brunope, line 176: these multi-line inline if statements are kinda weirdin me out, sure it doesn't need a backslash or somethnig?

ryantimjohn: I thought so too but flake8 says this is where they go, I got a hanging indent error for trying to make them look reasonable.

ryantimjohn: I can think about just putting everything in parens

Brunope:
Generally, all the `cast`ing is confusing to me, I know the code was already using it though. Why is it necessary? Is it because `DBDriver` and `RecordsSchemaField` don't use type hints so we have to cast their fields for them? To me, one of the big advantages of type-checking is a compile-time guarantee that you're passing all the right types to all the right functions (for us when we `make typecheck`). But type casting voids that guarantee cause you can type cast to whatever type want, so why even have the hint in the first place?

Anyway, just musing on the general state of this project, nothing for us to do about this right now

Brunope (approved):

ryantimjohn:
> Generally, all the `cast`ing is confusing to me, I know the code was already using it though. Why is it necessary? Is it because `DBDriver` and `RecordsSchemaField` don't use type hints so we have to cast their fields for them? To me, one of the big advantages of type-checking is a compile-time guarantee that you're passing all the right types to all the right functions (for us when we `make typecheck`). But type casting voids that guarantee cause you can type cast to whatever type want, so why even have the hint in the first place?
> 
> Anyway, just musing on the general state of this project, nothing for us to do about this right now

The reason we're using type hints here in particular is that lambdas don't allow type hinting and casting works as a workaround so that we pass mypy typechecking but, yes, you're right, casting != type hinting so we wouldn't get the compile time error that we're looking for. I guess it's a good reason to use fully scoped functions?

PR #225, opened 2023-03-10
Kill sphinx_requirements.txt - readthedocs build now uses setup.py

Brunope:


ryantimjohn (approved):
Looking over #226 you might have some merge work to do but I'll trust you to handle that.

PR #227, opened 2023-03-15
RM-46 simplify redshift records_unload

ryantimjohn:


records_mover/db/redshift/records_unload.py
```diff
@@ -10,6 +10,27 @@
...
+            cant_handle_hint(fail_if_cant_handle_hint, 'quotechar', hints)
+        redshift_options['add_quotes'] = True
+    elif hints.quoting is None:
+        redshift_options['add_quotes'] = False
+    else:
+        cant_handle_hint(fail_if_cant_handle_hint, 'quoting', hints)
+    quiet_remove(unhandled_hints, 'quoting')
+    quiet_remove(unhandled_hints, 'doublequote')
+    quiet_remove(unhandled_hints, 'quotechar')
+    return redshift_options
```
Brunope, line 31: Aren't you modifying the dict in-place already? no need to return it

ryantimjohn: Definitely, I always have a tough time remembering in Python that when you're not passing around base data types you're passing pointers to the objects rather than copies of the objects.

ryantimjohn: Probably because pointers is the wrong way to think about it but... mental models die hard. I was a C guy =(

Brunope: Well python is written in C so it's definitely all pointers under the hood

ryantimjohn: tru tru

ryantimjohn: It's just that like... it makes more sense when I think of python as names and objects rather than pointers and dereferencing pointers. 🤷

Brunope (approved):
looks good!

PR #232, opened 2023-03-24
RM-69 Remove bigfiles check

Brunope:


Brunope:
Tried adding extra dummy lines to setup.py and it didn't trigger a quality failure after removing the bigfiles check, so I'd say this is working.

ryantimjohn (approved):

PR #233, opened 2023-03-27
RM-48 factor to_pandas_dtype into dictionary

ryantimjohn:


records_mover/records/schema/field/__init__.py
```diff
@@ -199,96 +199,104 @@ def components_to_time_str(df: pd.DataFrame) -> datetime.time:
...
+            'string': np.object_,
+            'date': np.object_,
+            'datetime': 'datetime64[ns]',
+            'datetimetz': 'datetime64[ns, UTC]',
+            'time': np.object_,
+            'timetz': np.object_,
+        }
+
+        pd_dtype = field_type_to_pd_dtype_map.get(self.field_type)
+        if not pd_dtype:
```
Brunope, line 293: `if pd_dtype is None`

records_mover/records/schema/field/__init__.py
```diff
@@ -199,96 +199,104 @@ def components_to_time_str(df: pd.DataFrame) -> datetime.time:
...
             raise NotImplementedError("Teach me how to handle records schema "
                                       f"type {self.field_type}")
+        if self.field_type in ('integer', 'decimal') and callable(pd_dtype):
```
Brunope, line 296: kind of weird to make the map but then have to special case certain types again anyway.

Brunope (approved):

records_mover/records/schema/field/__init__.py
ryantimjohn: Yeah... the other option was to have everything be a lambda again which was kind of a drag. Open to talking through ideas here on Monday!

PR #236, opened 2023-04-04
RM-72 update version to 1.5.3

Brunope:


Brunope:
What is up with these failures? They look like real failures, but I can't reproduce them locally, and they only occur for the github actions unit tests, not the circleci ones

Brunope:
Ah, turns out pandas just released 2.0 and our github runners were installing that

ryantimjohn (approved):
LGTM!

PR #237, opened 2023-04-10
RM-45-records_mover-db-redshift-records_copy.py-13-1-C901-redshift_copy_options-is-too-complex-27

ryantimjohn:


records_mover/records/targets/fileobj.py
```diff
@@ -74,8 +74,7 @@ def write_dfs(path_or_buf: Union[str, IO[str]]) -> int:
             move_count = write_dfs(text_fileobj)
             text_fileobj.detach()
         else:
-            with NamedTemporaryFile(prefix='mover_fileobj_target') as output_file:
-                move_count = write_dfs(output_file.name)
```
Brunope, line 78: here's the reason your unit tests broken

ryantimjohn: ty!

records_mover/db/redshift/records_copy.py
```diff
@@ -3,92 +3,121 @@
...
-                          max_failure_rows: Optional[int]) -> RedshiftCopyOptions:
-    redshift_options: RedshiftCopyOptions = {}
+class RedshiftCopyParamaters:
+    def __init__(self,
+                 records_format: DelimitedRecordsFormat,
+                 fail_if_cant_handle_hint: bool,
+                 unhandled_hints: Set[str],
+                 fail_if_row_invalid: bool,
+                 max_failure_rows: Optional[int]):
+        self.redshift_options: RedshiftCopyOptions = dict()
```
Brunope, line 21: having both `RedshiftCopyParameters` and `RedshiftCopyOptions` is kinda confusing to me

records_mover/db/redshift/records_copy.py
```diff
@@ -3,92 +3,121 @@
...
-    elif hints.compression == 'BZIP':
-        redshift_options['compression'] = Compression.bzip2
-    elif hints.compression is None:
-        redshift_options['compression'] = None
+def process_compression(redshift_copy_parameters: RedshiftCopyParamaters):
+    compression_map = {
+        'GZIP': Compression.gzip,
+        'LZO': Compression.lzop,
+        'BZIP': Compression.bzip2,
+        None: None
```
Brunope, line 35: Is this `None: None` necessary?

records_mover/db/redshift/records_copy.py
```diff
@@ -3,92 +3,121 @@
...
 RedshiftCopyOptions = Dict[str, Optional[object]]
 
 
-def redshift_copy_options(unhandled_hints: Set[str],
-                          records_format: BaseRecordsFormat,
-                          fail_if_cant_handle_hint: bool,
-                          fail_if_row_invalid: bool,
-                          max_failure_rows: Optional[int]) -> RedshiftCopyOptions:
-    redshift_options: RedshiftCopyOptions = {}
+class RedshiftCopyParamaters:
```
Brunope, line 14: What do you think of using a factory abstraction instead? It seems like what we want to do is initialize _something_ with a few settings (fail_if_cant_handle_hint, fail_if_row_invalid, etc), then use that object to also store some state (unhandled_hints, copy_options) as we generate the desired output params.

Basically I think having a top-level object that stores these as `self` vars and has the `process_x` functions as methods could make a lot of sense. Thoughts?

ryantimjohn:
converted back to job for the Factory pattern refactor

records_mover/db/redshift/records_copy.py
ryantimjohn: I actually don't know that I'm going to write this as a factory pattern but we can look at it afterward and see if we need to refactor again.

records_mover/db/redshift/records_copy.py
ryantimjohn: refactored

ryantimjohn:
@Brunope just added docstrings after the approval so assuming it's okay to merge 😉

PR #238, opened 2023-04-10
RM-74 ignore fileobject type for mypy typecheck

Brunope:


Brunope:
Mypy updated and can no longer "infer" the type args to `shutil.copyfileobj`, even though we've already type annotated them. I tried adding extra annotations with no luck, so I'm opting to just type-ignore it.

ryantimjohn (approved):
Lgtm

PR #239, opened 2023-04-17
RM-44 Wrap determine_input_date_order_style in a handler object

Brunope:
The handler stores state as it processes and removes the various hints, rather than computing it all in one big function.


Brunope:
When I run the `cicoverage` test locally it thinks I have higher coverage than what it gets in automation, and wants me to increase the watermark rather than decrease it, what's up with that?

metrics/mypy_high_water_mark
```diff
@@ -1 +1 @@
-90.6000
+89.7400
```
Brunope, line 1: I'm not sure why the coverage had to go down, to me it looks like everything is appropriately type hinted?

records_mover/db/postgres/copy_options/date_input_style.py
```diff
@@ -40,342 +31,377 @@ def upgrade_date_order_style(style: DateOrderStyle, hint_name: str) -> None:
...
+            #  postgres=# select timestamptz '01-02-1999 23:01:01-10';
+            #        timestamptz
+            #  ------------------------
+            #   1999-01-03 09:01:01+00
+            #  (1 row)
+            #
+            self.upgrade_date_order_style('MDY', 'datetimeformattz')
+        else:
+            cant_handle_hint(self.fail_if_cant_handle_hint,
+                             'datetimeformattz', self.hints)
```
ryantimjohn, line 160: @Brunope might it make more sense to use a dictionary for this, just having the dictionary pass back the `'MDY', 'datetimeformattz'` (or similar) as a tuple?

records_mover/db/postgres/copy_options/date_input_style.py
```diff
@@ -40,342 +31,377 @@ def upgrade_date_order_style(style: DateOrderStyle, hint_name: str) -> None:
...
-        #
-        #  postgres=#
+    def process_hint_datetimeformat(self):
+        # datetimeformat: Valid values: "YYYY-MM-DD HH24:MI:SS",
+        # "YYYY-MM-DD HH12:MI AM", "MM/DD/YY HH24:MI". See Redshift docs
+        # for more information.
+
+        datetimeformat = self.hints.datetimeformat
+
+        if datetimeformat in ("YYYY-MM-DD HH24:MI:SS",
```
ryantimjohn, line 169: Same here viz if/else vs dictionary; totally understand if you prefer the if/else but lets talk through on call today.

records_mover/db/postgres/copy_options/date_input_style.py
ryantimjohn: Sounds good!

ryantimjohn (approved):

PR #242, opened 2023-04-28
RM-77 Check if BigQuery timezone drop issue still present 

ryantimjohn:


ryantimjohn:
Unfortunately I'm not able to find a viable workaround for this; it's become extremely difficult to test, also because Bigquery's interface has diverged very much from its api.

PR #244, opened 2023-05-09
RM-86 Error on Future and Deprecation Warnings

ryantimjohn:
We now have a nightly build. We'd like that to fail on Future and Deprecation Warnings so that we correct them before they mess up someone's install.


pytest.ini
ryantimjohn, line 1: @Brunope especially for your visibility, this pytest.ini is what makes the pytest command error on Deprecation and Future warning

tests/integration/records/single_db/pytest.ini
ryantimjohn, line 1: @Brunope especially for your visibility, this pytest.ini is what makes the pytest command error on Deprecation and Future warning

Brunope (approved):

PR #245, opened 2023-05-17
Rm 87 upgrade to sql alchemy 2.0

ryantimjohn:


tests/integration/records/multi_db/test_dataframe_schema_sql_creation.py
```diff
@@ -1,10 +1,10 @@
...
 from records_mover.records.processing_instructions import ProcessingInstructions
-from sqlalchemy_redshift.dialect import RedshiftDialect
-import unittest
-from unittest.mock import Mock
-from pandas import DataFrame
+from records_mover import Session, set_stream_logging
+import logging
 
 
 class TestDataframeSchemaSqlCreation(unittest.TestCase):
```
Brunope, line 10: nice, i like that this test changed to a real test without mocks

records_mover/db/redshift/redshift_db_driver.py
```diff
@@ -66,7 +66,7 @@ def table(self, schema: str, table: str) -> Table:
                 # There was some hang-wringing when they added it originally:
                 #  https://github.com/sqlalchemy-redshift/sqlalchemy-redshift/commit/
                 #      94723ec6437c5e5197fcf785845499e81640b167
-                return Table(table, self.meta, schema=schema, autoload=True, autoload_with=conn)
+                return Table(table, self.meta, schema=schema, autoload_with=conn)
```
Brunope, line 69: wait, now im confused. Earlier we passed `autoload_with=self.db_engine`, but now we are passing the connection? i thought using the engine/connection as a union was gone in 2.0?

records_mover/db/vertica/unloader.py
```diff
@@ -97,7 +101,13 @@ def unload_to_s3_directory(self,
         processing_instructions = unload_plan.processing_instructions
         try:
             s3_sql = self.aws_creds_sql(aws_creds.access_key, aws_creds.secret_key)
-            self.db.execute(s3_sql)
+            if isinstance(self.db, sqlalchemy.engine.Connection):
```
Brunope, line 104: eugh, i hate that this object is constructed with either an engine or connection, and we don't know which. would it be a lot of work to fix that?

records_mover/records/prep.py
```diff
@@ -35,7 +36,7 @@ def add_permissions(self, conn: Connection, driver: DBDriver) -> None:
 
     def create_table(self, schema_sql: str, conn: Connection, driver: DBDriver) -> None:
         logger.info('Creating table...')
-        conn.execute(schema_sql)
+        conn.exec_driver_sql(schema_sql)  # type: ignore
```
Brunope, line 39: how come doing this here instead of elsewhere doing `execute(text(sql))`?

records_mover/records/prep.py
```diff
@@ -55,11 +56,11 @@ def prep_table_for_load(self,
                                                            self.tbl.table_name)
             if (how_to_prep == ExistingTableHandling.TRUNCATE_AND_OVERWRITE):
                 logger.info("Truncating...")
-                db.execute(f"TRUNCATE TABLE {schema_and_table}")
+                db.execute(text(f"TRUNCATE TABLE {schema_and_table}"))
```
Brunope, line 59: also im confused about using the `db` object to do operations, shouldn't these be using the connection instead?

setup.py
```diff
@@ -224,10 +224,7 @@ def initialize_options(self) -> None:
 ]
 
 vertica_dependencies = [
-    # sqlalchemy-vertica-python 0.5.5 introduced
-    # https://github.com/bluelabsio/sqlalchemy-vertica-python/pull/7
-    # which fixed a bug pulling schema information from Vertica
-    'sqlalchemy-vertica-python>=0.5.5,<0.6',
+    'sqlalchemy-vertica-python',
```
Brunope, line 227: do we still want to keep the min version?

tests/integration/records/mover_test_case.py
```diff
@@ -27,7 +27,7 @@ def database_has_no_usable_timestamptz_type(self, engine: Engine) -> bool:
         # timezone is stripped off without looking at it before the
         # timestamp itself is stored without any modifications to the
         # hour number.
-        return engine.name == 'mysql'
+        return engine.name == 'mysql' or engine.name is None
```
Brunope, line 30: can the engine itself ever be none? what causes only the name to be none?

tests/integration/records/records_numeric_database_fixture.py
```diff
@@ -126,14 +127,18 @@ def bring_up(self):
...
+                    connection.exec_driver_sql(statement)
 
     def quote_schema_and_table(self, schema, table):
         return quote_schema_and_table(self.engine, schema, table)
 
     def drop_table_if_exists(self, schema, table):
         sql = f"DROP TABLE IF EXISTS {self.quote_schema_and_table(schema, table)}"
-        self.engine.execute(sql)
+        with self.engine.connect() as connection:
+            with connection.begin():
```
Brunope, line 140: are you sure you need to start a transaction to drop table?

tests/integration/records/single_db/test_records_load_datetime.py
```diff
@@ -58,12 +59,16 @@ def load(self,
...
     def pull_result(self,
                     column_name: str) -> datetime.datetime:
-        out = self.engine.execute(f'SELECT {column_name} '
-                                  f'from {self.schema_name}.{self.table_name}')
-        ret_all = out.fetchall()
+        with self.engine.connect() as connection:
+            with connection.begin():
+                out = connection.execute(text(
+                        f'SELECT {column_name} '
+                        f'from {self.schema_name}.{self.table_name}'))
```
Brunope, line 66: I think you should be using `bindparams` or something for variable substitutions. In general using format strings to generate SQL is a big no-no because of injection vulnerabilities.

https://docs.sqlalchemy.org/en/20/core/sqlelement.html#sqlalchemy.sql.expression.TextClause.bindparams

I know the code was already like this and it's not your fault, so I'm okay with doing this as followup work but I think it's important and I don't want us to forget about it

tests/integration/records/table_timezone_validator.py
```diff
@@ -135,7 +135,7 @@ def validate(self,
 
         if self.tc.raw_avro_types_written():
             assert timestamptzstr == '946834496789012'
-        elif (self.source_db_engine is not None and
+        elif (self.source_db_engine is None or
```
Brunope, line 138: why did this change from `not None` to `None`? that looks wrong

tests/integration/records/timezone.py
```diff
@@ -13,5 +15,5 @@ def set_session_tz(conn):
     # being of the session timezone.
     if conn.engine.name == 'vertica':
         tz = DB_SESSION_TIMEZONE
-        conn.execute('SET TIME ZONE TO :tz', tz=tz)
+        conn.execute(text('SET TIME ZONE TO :tz'), {'tz': tz})
```
Brunope, line 18: yeahhhh this is how you should do the format strings!

tests/unit/db/redshift/test_redshift_db_driver_unload.py
```diff
@@ -47,7 +47,8 @@ def test_unload_to_non_s3(self,
             'unload_location': self.mock_s3_temp_base_loc.temporary_directory().__enter__().url
         }
         mock_UnloadFromSelect.assert_called_with(**expected_args)
-        self.assertEqual(456, rows)
+        self.assertEqual('db_engine.connect().__enter__().execute().scalar()',
+                         rows._extract_mock_name())
```
Brunope, line 51: having a hard time understanding what this is doing, can you explain?

tests/unit/db/redshift/test_sql.py
```diff
@@ -12,11 +12,5 @@ def test_schema_sql_from_admin_views_not_installed(self,
...
-        out = schema_sql_from_admin_views(mock_schema, mock_table, mock_db)
-        self.assertIsNone(out)
-        mock_logger.debug.assert_called_with('Error while generating SQL', exc_info=True)
-        mock_logger.warning.\
-            assert_called_with("To be able to save SQL to a records directory, "
-                               "please install and grant access to 'admin.v_generate_tbl_ddl' "
-                               "from https://github.com/awslabs/amazon-redshift-utils/"
-                               "blob/master/src/AdminViews/v_generate_tbl_ddl.sql")
+        with self.assertRaises(AttributeError):
+            schema_sql_from_admin_views(mock_schema, mock_table, mock_db)
```
Brunope, line 16: that logging error seems useful.. does it still get logged or is the new behavior simply an attribute error?

tests/unit/db/vertica/test_unloader_no_aws_creds.py
```diff
@@ -49,7 +49,7 @@ def test_unload_with_no_aws_creds(self,
 
         mock_column = Mock(name='column')
         mock_db.dialect.get_columns.return_value = [mock_column]
-        with self.assertRaises(NotImplementedError):
+        with self.assertRaises(AttributeError):
```
Brunope, line 52: where is this behavior changing and why? is the attribute error an internal thing from sqlalchemy getting surfaced?

tests/unit/db/vertica/test_vertica_db_driver.py
```diff
@@ -14,13 +14,12 @@ def test_unload(self):
...
         self.mock_records_unload_plan.records_format.hints = vertica_format_hints
         self.mock_directory.scheme = 's3'
-        export_count = self.vertica_db_driver\
-            .unloader().unload(schema='myschema',
-                               table='mytable',
-                               unload_plan=self.mock_records_unload_plan,
-                               directory=self.mock_directory)
-
-        self.assertEqual(579, export_count)
+        with self.assertRaises(NotImplementedError):
```
Brunope, line 17: how come this is back to `NotImplementedError` and not `AttributeError` like the other cases?

ryantimjohn:
@Brunope to address some of your concerns, I'm going to have to do a somewhat more thorough rewrite and actually remove all instances of the Union[engine, connection] that are so common throughout the package. Overall, it'll be much better for it so am going to get on this.

records_mover/db/vertica/unloader.py
ryantimjohn: Fixed

records_mover/records/prep.py
ryantimjohn: I think I might have to do a rewrite of the whole driver interface to get this working; this is going to be a pain.

setup.py
ryantimjohn: reinstated min version

tests/integration/records/records_numeric_database_fixture.py
ryantimjohn: nope!

tests/integration/records/single_db/test_records_load_datetime.py
```diff
@@ -58,12 +59,16 @@ def load(self,
 
     def pull_result(self,
                     column_name: str) -> datetime.datetime:
-        out = self.engine.execute(f'SELECT {column_name} '
-                                  f'from {self.schema_name}.{self.table_name}')
-        ret_all = out.fetchall()
+        with self.engine.connect() as connection:
+            with connection.begin():
```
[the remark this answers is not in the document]
ryantimjohn: nope!

tests/integration/records/single_db/test_records_load_datetime.py
ryantimjohn: I created this ticket so we can go in and try and clean out all SQL injection possibilities:
https://bluelabsio.atlassian.net/browse/RM-95

tests/integration/records/records_numeric_database_fixture.py
ryantimjohn: so... can't say why but I guess we do need this begin

tests/integration/records/single_db/test_records_load_datetime.py
ryantimjohn: Unfortunately using bind_params or passing a dictionary of params is causing errors in the code. It seems like this isn't meant for passing table names, according to my reading. I think I'm going to investigate just using table constructs for all of these queries, like so: https://docs.sqlalchemy.org/en/20/core/selectable.html#sqlalchemy.sql.expression.table

records_mover/records/prep.py
ryantimjohn: https://bluelabsio.atlassian.net/jira/software/projects/RM/boards/3?selectedIssue=RM-97
ticket here, I'm going to address this in a future PR and not this one if that's alright @Brunope

tests/integration/records/single_db/test_records_load_datetime.py
ryantimjohn: Going to address this in a separate ticket here @Brunope:
https://bluelabsio.atlassian.net/jira/software/projects/RM/boards/3?selectedIssue=RM-95

tests/integration/records/table_timezone_validator.py
ryantimjohn: After the switch to SQLAlchemy 2.0, when we have a `None` db_engine, the output is now:
`2000-01-02 17:34:16.120786 ` (for example) whereas it used to be `2000-01-02 17:34:16.120786+00`. I was looking into this but I don't have a good reason why this changed.

tests/integration/records/timezone.py
ryantimjohn: So this works for real parameters, but doesn't work for table or schema names (probably on purpose on SQLAlchemy's part, they want you to use the much safer `Table` object.

tests/unit/db/redshift/test_redshift_db_driver_unload.py
ryantimjohn: This is now fixed

tests/unit/db/redshift/test_sql.py
ryantimjohn: This now calls correctly by using MagicMocks

tests/unit/db/vertica/test_unloader_no_aws_creds.py
ryantimjohn: Fixed this, now using a MagicMock to reproduce NotImplementedError

tests/unit/db/vertica/test_unloader_no_aws_creds.py
```diff
@@ -63,4 +63,8 @@ def test_s3_export_available_false_no_awslib(self):
         mock_out = mock_db.execute.return_value
         mock_out.fetchall.return_value = []
         unloader = VerticaUnloader(db=mock_db, s3_temp_base_loc=mock_s3_temp_base_loc)
-        self.assertEqual(False, unloader.s3_export_available())
+        try:
+            s3_export_available = unloader.s3_export_available()
```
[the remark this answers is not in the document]
ryantimjohn: Absolutely, resolved.

ryantimjohn:
> We need an integration test that demonstrates old sqlalchemy continues to work, especially for when we move forward and continue updating sqlalchemy versions and related pieces of code.

This is done now, both as an integration test with all DBs and as a unit test on Python 3.11 (thought this would make it long-lived as a test.)

tests/integration/records/single_db/test_records_load_datetime.py
ryantimjohn: Calling this resolved as we're going to address SQL Injection in the other ticket.

records_mover/records/prep.py
ryantimjohn: Okay, this is now resolved. All calls are to either db_engine or db_conn depending on what the action is. We've completely separated out these interfaces now.

ryantimjohn:
> We need an integration test that demonstrates old sqlalchemy continues to work, especially for when we move forward and continue updating sqlalchemy versions and related pieces of code.

This is now part of our testing:
https://github.com/bluelabsio/records-mover/blob/8ba646e0f31b02a7c6fcdf660a3f8cb31f96f457/.circleci/config.yml#L505C8-L514

tests/integration/records/single_db/test_records_load_datetime.py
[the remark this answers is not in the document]
ryantimjohn: Turns out that it was; SQLAlchemy now doesn't let you execute without explicitly beginning a transaction.

ryantimjohn:
Alright @Brunope this is ready for your eyes again

records_mover/airflow/hooks/records_hook.py
```diff
@@ -50,12 +51,15 @@ def _url_resolver(self) -> UrlResolver:
                            gcs_client_getter=lambda: None,
                            gcp_credentials_getter=lambda: None)
 
-    def _db_driver(self, db: Union[sqlalchemy.engine.Engine,
-                                   sqlalchemy.engine.Connection]) -> DBDriver:
+    def _db_driver(self, db: Optional[Union[sqlalchemy.engine.Engine,
```
Brunope, line 54: is this supposed to have `@property` like `_url_resolver` above does?

records_mover/check_db_conn_engine.py
```diff
@@ -0,0 +1,31 @@
+# flake8: noqa
+
+import sqlalchemy
+from typing import Union, Optional, Tuple
+import warnings
+
+
+def check_db_conn_engine(db: Optional[Union[sqlalchemy.engine.Engine,
```
Brunope, line 8: this is a nicely encapsulated function that could definitely benefit from unit tests for each of its branches

records_mover/check_db_conn_engine.py
```diff
@@ -0,0 +1,31 @@
...
+                      "objects.",
+                      DeprecationWarning)
+    if not (db or db_conn or db_engine):
+        raise ValueError("Either db, db_conn, or db_engine must be provided as arguments")
+    if isinstance(db, sqlalchemy.engine.Connection) and not db_conn:
+        db_conn = db
+    if isinstance(db, sqlalchemy.engine.Engine) and not db_engine:
+        db_engine = db
+    if not db_engine:
+        print("db_engine is not provided, so we're creating one from db_conn")
```
Brunope, line 29: 'creating' maybe isnt the best word, its more like taking a reference from i think

ryantimjohn: great point, updated!

records_mover/airflow/hooks/records_hook.py
ryantimjohn: I don't think so, strangely enough _db_driver seems to be a function and not a property. I'm going to check that I'm right on this.

ryantimjohn: I think I am, this is very awkward syntax but: it seems like it's not actually a property of the class but that this is instantiating a db_driver through the db_driver factory so I don't think this would be appropriate to assign to the "getter" method of the @property _db_driver. It's possible that it could be assigned to the setter method for the property _db_driver property but that might have downstream effects that I don't want.

I agree that this wasn't really the best method name (should have been create_db_driver or something like that) but I'm hesitant to change it at this point because it gets called all over the place.

records_mover/check_db_conn_engine.py
ryantimjohn: Great point, added here: https://github.com/bluelabsio/records-mover/blob/f22acfd664b99283ba8dadf07528aabf5d93bf17/tests/unit/test_check_db_conn_engine.py

records_mover/airflow/hooks/records_hook.py
ryantimjohn: Let's talk through this at our meeting Monday just to make sure that I'm not lost in my thinking.

ryantimjohn: Resolved per our chat

records_mover/db/bigquery/loader.py
```diff
@@ -19,19 +19,41 @@
...
+                 db_engine: Optional[sqlalchemy.engine.Engine] = None) -> None:
+        db, db_conn, db_engine = check_db_conn_engine(db=db, db_conn=db_conn, db_engine=db_engine)
         self.db = db
+        self._db_conn = db_conn
+        self.db_engine = db_engine
         self.url_resolver = url_resolver
         self.gcs_temp_base_loc = gcs_temp_base_loc
+        self.conn_opened_here = False
+
+    def get_db_conn(self) -> sqlalchemy.engine.Connection:
```
Brunope, line 42: I see a lot of loaders that have this same pattern of a `get_db_conn` method and setter that does the same thing. Have you thought about extracting this into some shared base object?

Personally, I'm okay with limiting the amount of logic dumped into inherited base classes because I find the resulting code harder to follow, and this is relatively clear. 

But I wanted to bring it up as an option to be considered

Brunope: One advantage of extracting the logic could be when we move to change the db/dbconn/dbengine behavior again, it might make it simpler to update and test if its in fewer places

records_mover/db/bigquery/loader.py
```diff
@@ -19,19 +19,41 @@
...
+            self.conn_opened_here = True
+            logger.debug(f"Opened connection to database within {self} because none was provided.")
+        return self._db_conn
+
+    def set_db_conn(self, db_conn: Optional[sqlalchemy.engine.Connection]) -> None:
+        self._db_conn = db_conn
+
+    def del_db_conn(self) -> None:
+        if self.conn_opened_here:
+            self.db_conn.close()
```
Brunope, line 54: I think you should also set `self.db_conn = None`

records_mover/db/driver.py
```diff
@@ -42,43 +65,63 @@ def schema_sql(self,
...
+        return str(CreateTable(table_obj))
 
     def varchar_length_is_in_chars(self) -> bool:
         """True if the 'n' in VARCHAR(n) is represented in natural language
         characters, rather than in post-encoding bytes.  This varies by database -
         override it to control"""
         return False
 
-    def set_grant_permissions_for_groups(self, schema_name: str, table: str,
+    def set_grant_permissions_for_groups(self,
```
Brunope, line 76: we probably shouldnt change this but its really weird to me that this method takse a `db` as input, even though the constructor of the dbdriver already does this.

records_mover/db/driver.py
```diff
@@ -42,43 +65,63 @@ def schema_sql(self,
...
                 if not perm_type.isalpha():
                     raise TypeError("Please make sure your permission types"
                                     " are an acceptable value.")
                 perms_sql = f'GRANT {perm_type} ON TABLE {schema_and_table} TO {group_name}'
-                db.execute(perms_sql)
+                if db_conn:
+                    db_conn.execute(perms_sql)
+                else:
+                    with db_engine.connect() as conn:
+                        conn.execute(perms_sql)
```
Brunope, line 100: would be nice to be able to leverage `self.get_db_conn()` for this, but i see that the db we have here isn't `self` so i  don't think we can. eugh

records_mover/db/driver.py
Brunope: is there a reason for this? why can't we just use the existing db stuff in `self`? is there a use case where we need a different db on the same driver object?

records_mover/db/mysql/mysql_db_driver.py
```diff
@@ -17,11 +17,16 @@
...
 class MySQLDBDriver(DBDriver):
     def __init__(self,
-                 db: Union[sqlalchemy.engine.Engine, sqlalchemy.engine.Connection],
                  url_resolver: UrlResolver,
+                 db: Optional[Union[sqlalchemy.engine.Engine, sqlalchemy.engine.Connection]] = None,
+                 db_conn: Optional[sqlalchemy.engine.Connection] = None,
+                 db_engine: Optional[sqlalchemy.engine.Engine] = None,
                  **kwargs) -> None:
-        super().__init__(db)
+        self.conn_opened_here = False
```
Brunope, line 25: why does this class set `conn_opened_here` but `bigquery_db_driver` does not?

records_mover/db/bigquery/loader.py
ryantimjohn: Totally agree but, to your above point, I really try and shy away for very abstract inheritance as much as possible and this wasn't a place where I saw an easy way to implement a good inheritance pattern.

I always try to go with the `composition better than inheritance` mindset but since we're dealing with a property here, I didn't see an easy way to do that but happy to try and brainstorm with you!

ryantimjohn: Alright @Brunope what do you think about this:
https://github.com/bluelabsio/records-mover/pull/255

records_mover/db/bigquery/loader.py
ryantimjohn: to trigger garbage collection/del method? sounds good.

records_mover/db/db_conn_composable_methods.py
ryantimjohn, line 1: Note to add tests for this file

records_mover/db/driver.py
ryantimjohn: I can't think of a reason that it would take a different db, unless for some reason this db_driver needed to set permissions for a db it was writing to, but I don't see why that wouldn't be done by that db's db_driver. I agree this is super bizarre, I also agree that it'd be maybe more of a pain than it's worth to change.

records_mover/db/driver.py
ryantimjohn: yeah, was thinking the same thing but agree that we're just going to have to go with whatever db is passed

records_mover/db/mysql/mysql_db_driver.py
ryantimjohn: oversight. The conn_opened_here gets set in the super().__init__

ryantimjohn: resolved!

records_mover/db/db_conn_mixin.py
```diff
@@ -0,0 +1,38 @@
+import logging
```
Brunope, line 1: i like this mixin. I feel like somewhere in the code this PR adds we should add a comment explaining our general constraints of wanting to not break existing python code that might have relied on this "db" property. Which greatly shaped the structure and changes of the code you wrote. I imagine someone reading this code in the future without realizing this constraint would be very confused about why some things are the way they are now.

Maybe somewhere in this class is the right place to add such a comment? Maybe near `check_db_conn_engine`? not sure

records_mover/db/bigquery/bigquery_db_driver.py
```diff
@@ -19,17 +19,23 @@
 
 class BigQueryDBDriver(DBDriver):
     def __init__(self,
-                 db: Union[sqlalchemy.engine.Connection, sqlalchemy.engine.Engine],
+                 db: Optional[Union[sqlalchemy.engine.Connection, sqlalchemy.engine.Engine]],
```
Brunope, line 22: why does the BigQueryLoader use the mixin but the BigQueryDriver doesnt?

Brunope: Is it because in the loader itself it might have to open a connection so it cares more about that sort of state?

records_mover/check_db_conn_engine.py
```diff
@@ -0,0 +1,31 @@
...
+    if db:
+        warnings.warn("The db argument is deprecated and will be"
+                      "removed in future releases.\n"
+                      "Please use db_conn for Connection objects and db_engine for Engine"
+                      "objects.",
+                      DeprecationWarning)
+    if not (db or db_conn or db_engine):
+        raise ValueError("Either db, db_conn, or db_engine must be provided as arguments")
+    if isinstance(db, sqlalchemy.engine.Connection) and not db_conn:
+        db_conn = db
```
Brunope, line 25: does the connection have an `engine` attribute or getter? I'm wondering if it would be useful in the case where we're passed in a connection to also return its corresponding engine alongside the connection instead of None for the engine.

If it does have such an attribute i think we could add the guarantee that the returned engine is never none which could be handy for some callers maybe? deferring to you on this but an idea

records_mover/db/factory.py
```diff
@@ -1,33 +1,38 @@
...
-                        sqlalchemy.engine.Connection],
+def db_driver(db: Optional[Union[sqlalchemy.engine.Engine,
+                           sqlalchemy.engine.Connection]],
+              db_conn: Optional[sqlalchemy.engine.Connection] = None,
+              db_engine: Optional[sqlalchemy.engine.Engine] = None,
               **kwargs) -> DBDriver:
-    engine: sqlalchemy.engine.Engine = db.engine
-    engine_name: str = engine.name
+    db, db_conn, db_engine = check_db_conn_engine(db=db, db_conn=db_conn, db_engine=db_engine)
+    engine_name: str = db_engine.name
```
Brunope, line 15: like for instance here, if `db` is a connection not an engine, wouldn't the returned engine be None and this would crash?

records_mover/db/postgres/loader.py
```diff
@@ -58,32 +67,30 @@ def load_from_fileobjs(self,
...
+        # https://www.postgresql.org/docs/8.3/sql-set.html
+        #
+        # The effects of SET LOCAL last only till the end of the
+        # current transaction, whether committed or not. A special
+        # case is SET followed by SET LOCAL within a single
+        # transaction: the SET LOCAL value will be seen until the end
+        # of the transaction, but afterwards (if the transaction is
+        # committed) the SET value will take effect.
+        date_style = f"ISO, {date_order_style}"
+        sql = f"SET LOCAL DateStyle = {quote_value(None, date_style, db_engine=self.db_engine)}"
```
Brunope, line 81: where does date_order_style come from? Shouldn't this be using the proper parameter substitution? I know you're quoting it so its better than nothing but im not an expert so i dont know if that's guaranteed to be safe still.

records_mover/db/postgres/loader.py
```diff
@@ -58,32 +67,30 @@ def load_from_fileobjs(self,
...
+        logger.info(sql)
+        self.db_conn.execute(text(sql))
```
Brunope, line 83: correct me if i am not understanding this right - so when we reference `self.db_conn` here, since it's None we'll create a new connection, and the opened_here flag will be set to True, because this class inherits from the mixin.

One question I have is is that gonna give the same transaction guarantees as the way the previous code was written? I know if you use a context manager like `with connection.begin()` it starts a transaction - but i don't think we are doing that here anymore so is that a problem??

records_mover/db/postgres/sqlalchemy_postgres_copy.pyi
```diff
@@ -1,17 +1,18 @@
-from typing import Union, IO
+from typing import Union, IO, Optional
 import sqlalchemy
 
 
 def copy_from(source: IO[bytes],
               dest: sqlalchemy.schema.Table,
-              engine_or_conn: Union[sqlalchemy.engine.Engine, sqlalchemy.engine.Connection],
+              engine_or_conn: Optional[Union[sqlalchemy.engine.Engine,
+                                             sqlalchemy.engine.Connection]],
```
Brunope, line 8: why do these change to optional? How are you going to copy if you dont have a connection or an engine? I know in some of the upper code that stores these objects theyre passed around as optional, but shouldnt you have to unwrap the option at some point before you actually do anything with it?

records_mover/db/postgres/unloader.py
```diff
@@ -43,30 +44,28 @@ def unload(self,
...
-                        fileobj,
-                        conn,
-                        **postgres_options)
+        filename = unload_plan.records_format.generate_filename('data')
+        loc = directory.loc.file_in_this_directory(filename)
+        with loc.open(mode='wb') as fileobj:
+            copy_to(table_obj.select(),
+                    fileobj,
+                    self.db_conn,
+                    **postgres_options)
```
Brunope, line 68: same transaction concerns as my previous comment

records_mover/db/vertica/export_sql.py
```diff
@@ -21,7 +21,7 @@ def vertica_export_sql(db_engine: Engine,
     """
 
     def quote(value: str) -> str:
-        return quote_value(db_engine, value)
+        return quote_value(None, value, db_engine=db_engine)
```
Brunope, line 24: im assuming these None values are the old `db`

records_mover/db/vertica/unloader.py
```diff
@@ -97,7 +103,11 @@ def unload_to_s3_directory(self,
         processing_instructions = unload_plan.processing_instructions
         try:
             s3_sql = self.aws_creds_sql(aws_creds.access_key, aws_creds.secret_key)
-            self.db.execute(s3_sql)
+            if not self.db_conn.in_transaction():
+                with self.db_conn.begin():
+                    self.db_conn.execute(text(s3_sql))
+            else:
+                self.db_conn.execute(text(s3_sql))
```
Brunope, line 110: yeah this kind of confusion is the downside of the approach we took. where you have to check the state of the passed in connection rather than explicitly always managing it yourself.  But once we move past this interim of still trying to support the old ways it should get better i think

records_mover/records/prep_and_load.py
```diff
@@ -16,24 +16,24 @@ def prep_and_load(tbl: TargetTableDetails,
                   load_exception_type: Type[Exception],
                   reset_before_reload: Callable[[], None] = lambda: None) -> MoveResult:
     logger.info("Connecting to database...")
-    with tbl.db_engine.begin() as db:
-        driver = tbl.db_driver(db)
+    with tbl.db_engine.begin() as db_conn:
```
Brunope, line 19: can you do this? db_engine.begin() does it create a connection and start a transaction all at once?

Brunope: i think it would be better to get the connection from the engine, then use `conn.begin()` the same way the majority if the rest of the code does  it - our overarching goal is treating engines like engines and connections like connections and not mixing them even if sqlalchemy allows us to.

records_mover/records/schema/schema/known_representation.py
```diff
@@ -60,7 +60,7 @@ def from_dataframe(df: 'DataFrame',
     def from_db_driver(driver: 'DBDriver',
                        schema_name: str,
                        table_name: str) -> 'RecordsSchemaSqlKnownRepresentation':
-        type = f"sql/{driver.db.dialect.name}"
+        type = f"sql/{driver.db_engine.dialect.name}"
```
Brunope, line 63: this is kinda nice, as a reader i can look at this code and reason, yeah an engine has a dialect with a name that we are referencing vs thinking wtf is this db object i hope its an engine lol

records_mover/records/sources/table.py
```diff
@@ -74,15 +75,11 @@ def to_dataframes_source(self,
...
+        db_engine = self.driver.db_engine
         records_schema = self.pull_records_schema()
 
-        if isinstance(db, Engine):
-            connection = db.connect()
-            columns = db.dialect.get_columns(connection, self.table_name, schema=self.schema_name)
-            connection.close()
-        else:
-            columns = db.dialect.get_columns(db, self.table_name, schema=self.schema_name)
+        columns = db_engine.dialect.get_columns(db_conn, self.table_name, schema=self.schema_name)
```
Brunope, line 82: this is a beautiful diff!

records_mover/records/targets/spectrum.py
```diff
@@ -70,14 +78,15 @@ def prep_bucket(self) -> None:
...
             if self.existing_table_handling == ExistingTableHandling.DROP_AND_RECREATE:
-                schema_and_table: str = quote_schema_and_table(self.db,
+                schema_and_table: str = quote_schema_and_table(None,
                                                                self.schema_name,
-                                                               self.table_name)
+                                                               self.table_name,
+                                                               db_engine=self.db_engine)
                 logger.info(f"Dropping external table {schema_and_table}...")
-                with self.db.connect() as cursor:
+                with self.db_engine.connect() as cursor:
```
Brunope, line 86: as `conn`, lets be consistent

records_mover/records/targets/table/move_from_fileobjs_source.py
```diff
@@ -53,8 +53,8 @@ def reset_before_reload(self) -> None:
         self.fileobj.seek(0)
 
     def move(self) -> MoveResult:
-        with self.tbl.db_engine.begin() as db:
-            driver = self.tbl.db_driver(db)
+        with self.tbl.db_engine.begin() as db_conn:
```
Brunope, line 56: same note here as my previous note on using `db_engine.begin()` vs first opening a connection and using `conn.begin()`

records_mover/records/targets/table/target.py
```diff
@@ -39,16 +39,20 @@ def __init__(self,
                  schema_name: str,
                  table_name: str,
                  db_engine: Engine,
-                 db_driver: Callable[[Union[Engine, Connection]], DBDriver],
+                 db_driver: Callable[[Optional[Union['Engine', 'Connection']],
+                                      Optional[Connection],
+                                      Optional[Engine]], DBDriver],
```
Brunope, line 44: its just such an ugly type declaration, can we use a typedef or alias or something to describe this factory parameter that isnt really a db_driver but returns one?

records_mover/utils/json_schema.py
```diff
@@ -52,7 +52,8 @@ def is_optional_type(tp: PythonType) -> bool:
 def is_impractical_type(python_type: PythonType) -> bool:
     # can't write code and pass it around in JSON!
     return (is_callable_type(python_type) or
-            type(python_type) == enum.EnumMeta)
+            type(python_type) == enum.EnumMeta
+            or type(python_type) == typing.ForwardRef)
```
Brunope, line 56: dont understand this change

tests/integration/records/multi_db/test_records_table2table.py
```diff
@@ -89,34 +105,18 @@ def move_and_verify(self, source_dbname: str, target_dbname: str) -> None:
...
-        quoted_target = quote_schema_and_table(target_engine, target_schema_name, TARGET_TABLE_NAME)
+        quoted_target = quote_schema_and_table(None, target_schema_name, TARGET_TABLE_NAME,
+                                               db_engine=target_engine)
         sql = f"DROP TABLE {quoted_target}"
-        target_engine.execute(sql)
+        target_conn.exec_driver_sql(sql)  # type: ignore[attr-defined]
 
-        records_database_fixture.tear_down()
+        # records_database_fixture.tear_down()
+        source_conn.close()
```
Brunope, line 114: should these be using `with` instead to close the connection? if the test fails before it gets to closing the connection are we gonna be leaving hannging connections around?

tests/integration/records/records_datetime_fixture.py
```diff
@@ -82,7 +96,13 @@ def createDateTimeTable(self) -> None:
 """  # noqa
         else:
             raise NotImplementedError(f"Please teach me how to integration test {self.engine.name}")
-        self.engine.execute(create_tables)
+        if not self.connection:
```
Brunope, line 99: why do we have to check this? is there a problem with always creating a new connection / transaction from our engine?

tests/unit/test_check_db_conn_engine.py
```diff
@@ -0,0 +1,84 @@
...
+        mock_db = MagicMock(name='db', spec=sqlalchemy.engine.Engine)
+        mock_db_conn = MagicMock(name='db_conn', spec=sqlalchemy.engine.Connection)
+        mock_db_engine = MagicMock(name='db_engine', spec=sqlalchemy.engine.Engine)
+        with self.assertWarns(DeprecationWarning):
+            db, db_conn, db_engine = check_db_conn_engine(db=mock_db,
+                                                          db_conn=mock_db_conn,
+                                                          db_engine=mock_db_engine)
+        self.assertEqual(mock_db, db)
+        self.assertEqual(mock_db_conn, db_conn)
+        self.assertEqual(mock_db_engine, db_engine)
```
Brunope, line 84: thumbsup

records_mover/db/db_conn_mixin.py
ryantimjohn: I think that's a great idea, will get on this

records_mover/db/bigquery/bigquery_db_driver.py
ryantimjohn: yeah, this is the problem with inconsistent use of hierarchies/inheritance as we mentioned (it would have been nice if all loaders, unloaders, etc. inherited from a base loader class but, here we are)

records_mover/db/factory.py
ryantimjohn: if db or db_conn is provided, and one of them is a connection, then db_engine would be assigned to db_conn.engine. Relevant lines of code linked:
https://github.com/bluelabsio/records-mover/blob/e5c006015b534cda1477c77ec5ec5da515437619/records_mover/check_db_conn_engine.py#L28-L30

I realize that's not obvious now, so I'm up for rewriting in a way that makes that more clear, maybe naming the functions check_and_assign_db_conn_engine or something like that. Tell me what you think!

records_mover/db/postgres/unloader.py
ryantimjohn: Same explanation, this is called in a wrapping method where `with db_conn.begin():` is used. Can try and break that out.

records_mover/db/vertica/export_sql.py
ryantimjohn: yep!

records_mover/db/factory.py
Brunope: oh yeah great i missed that but it was what i was envisioning!

records_mover/records/prep_and_load.py
ryantimjohn: This makes sense, easy rewrite! I was just trying to retain as much of the original writing as possible.

records_mover/records/schema/schema/known_representation.py
ryantimjohn: haha, def. The use of the term db throughout was just hella confusing on my first read throughs.

records_mover/records/targets/table/move_from_fileobjs_source.py
ryantimjohn: on it

records_mover/db/postgres/loader.py
Brunope: never mind then sounds good!

records_mover/records/targets/table/target.py
ryantimjohn: it's so annoying. I tried to create a Protocol for this but it didn't make it a lot clearer. I'm open to ideas but this is definitely a method and not a property so it has to be wrapped in this Callable type.

records_mover/utils/json_schema.py
ryantimjohn: Shoot, I should have documented it at the time because now I don't remember.

tests/integration/records/multi_db/test_records_table2table.py
ryantimjohn: I'm going to turn the records_database_fixture.tear_down() back on, I don't know why I turned this off but that definitely contains closure of the connection.

records_mover/check_db_conn_engine.py
ryantimjohn: We always pass back an engine!

records_mover/db/db_conn_mixin.py
ryantimjohn: Added an explainer in `check_db_conn_engine`

records_mover/db/postgres/sqlalchemy_postgres_copy.pyi
ryantimjohn: I see no reason for it to be Optional so removing that

records_mover/records/schema/schema/known_representation.py
ryantimjohn: Done!

records_mover/utils/json_schema.py
ryantimjohn: Okay, this is the error I get without the ForwardRef:
```
NotImplementedError: Teach me how to handle Python type ForwardRef('Connection') (parameter: db_conn)
```

ryantimjohn: So I do want to be able to pass connection strings around as far as I understand it, I'm not sure if this is new behavior or why I had to implement this now. I can delve deeper but if you're alright, I've just added an explainer as to why now

records_mover/records/targets/table/move_from_fileobjs_source.py
ryantimjohn: Updated these across the project (there were several)

Brunope (approved):

records_mover/records/targets/table/target.py
ryantimjohn: Alright, I tried again to write a protocol for this function but we run into a lot of problems with imports and such; I think I'm going to have to say that, while this is extremely ugly, it at least works so I'm going to have to leave as is. </3

tests/integration/records/multi_db/test_records_table2table.py
ryantimjohn: Used with syntax, brought back the tear_down

PR #253, opened 2023-08-28
RM-104: upper bound mypy < 1.5.0 to fix typechecking

Brunope:
https://github.com/python/mypy/issues/15891

installing mypy-1.4.1 fixed this issue

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -239,7 +239,7 @@ literally_every_single_database_binary_dependencies = (
 )
 
 typecheck_dependencies = [
-    'mypy',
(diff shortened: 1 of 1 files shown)


ryantimjohn (approved):
Looks good as a short term fix! Any easy way you can think of to track this ticket and lift the cap once it's fixed @Brunope ? Might just link to this Github Issue in a comment above the change:
https://github.com/python/mypy/issues/15891

ryantimjohn:
haha, so embarrassed, was looking at this half screen

PR #254, opened 2023-10-02
Rm 95 remove sql injection possibilities

ryantimjohn:


ryantimjohn:
@Brunope should be ready for review at the end of testing! (I just merged in main.)

records_mover/db/driver.py
```diff
@@ -68,16 +68,14 @@ def set_grant_permissions_for_groups(self,
...
-                                                       db_engine=self.db_engine)
         for perm_type in groups:
             groups_list = groups[perm_type]
             for group in groups_list:
-                group_name: str = quote_group_name(None, group, db_engine=self.db_engine)
                 if not perm_type.isalpha():
                     raise TypeError("Please make sure your permission types"
                                     " are an acceptable value.")
-                perms_sql = f'GRANT {perm_type} ON TABLE {schema_and_table} TO {group_name}'
+                table_obj = Table(table, MetaData(), schema=schema_name)
```
Brunope, line 77: clever

records_mover/db/driver.py
```diff
@@ -68,16 +68,14 @@ def set_grant_permissions_for_groups(self,
...
+                perms_sql = str(GrantPrivileges(perm_type, table_obj, group))
```
Brunope, line 78: should this be `text()` instead of `str()` or am i misunderstanding how you're supposed to you text()?

records_mover/db/mysql/load_options.py
```diff
@@ -69,10 +69,12 @@ def generate_load_data_sql(self,
                                filename: str,
                                schema_name: str,
                                table_name: str) -> TextClause:
+        remove_backticks_schema_name = schema_name.replace('`', '')
+        remove_backticks_table_name = table_name.replace('`', '')
         sql = f"""\
 LOAD DATA
 LOCAL INFILE :filename
-INTO TABLE {schema_name}.{table_name}
+INTO TABLE `{remove_backticks_schema_name}`.`{remove_backticks_table_name}`
```
Brunope, line 77: will this work?? if we passed a table originally containing backticks because we thought it sholud be quoted, it will end up as ``'table'`, is that equivalent to ``table`` without the single quotes?

records_mover/db/redshift/unloader.py
```diff
@@ -84,10 +84,9 @@ def unload_to_s3_directory(self,
             #
             register_secret(aws_creds.token)
             register_secret(aws_creds.secret_key)
-            select = text(
-                "SELECT * FROM "
-                f"{quote_schema_and_table(None, schema, table, db_engine=self.db_engine)}")
-            unload = UnloadFromSelect(select=select,
+            table_obj = Table(table, MetaData(), schema=schema)
+            select = str(sqlalchemy.select('*', table_obj))  # type: ignore[arg-type]  # noqa: F821
```
Brunope, line 88: i don't think you have to str() this if you are calling text() on it below right?

records_mover/records/prep.py
```diff
@@ -61,19 +62,24 @@ def prep_table_for_load(self,
                                                            db_engine=db_engine,)
             if (how_to_prep == ExistingTableHandling.TRUNCATE_AND_OVERWRITE):
                 logger.info("Truncating...")
-                db_conn.execute(text(f"TRUNCATE TABLE {schema_and_table}"))
+                meta = MetaData()
+                table = Table(self.tbl.table_name, meta, schema=self.tbl.schema_name)
```
Brunope, line 66: when constructing these table objs do you have to set the `autoload_with=engine` or anything for it to correctly initialize or no? ive always done that in the past not sure what its needed for

Brunope (approved):
had some minor concerns but will defer to you

records_mover/db/mysql/load_options.py
ryantimjohn: This is a great point, we also already have the quote schema and table method, going to try to use that

ryantimjohn: Actually, I can't use `quote_schema_and_table` here because I'd need to pass an engine. So, while this is a terrible solution, I think it's our best one.

records_mover/db/driver.py
ryantimjohn: need both because text expects a string or bytes-like object; here's the TypeError you get without converting first:
`TypeError: expected string or bytes-like object`

records_mover/db/redshift/unloader.py
ryantimjohn: As above, need to convert to string to pass to text

records_mover/records/prep.py
ryantimjohn: `autoload_with` is use to reflect the table:
https://docs.sqlalchemy.org/en/20/core/reflection.html

I don't think we actually need to do this with this statement because we're just deleting all rows from the table; reflection adds significant time overhead to running statements

PR #257, opened 2023-11-07
RM-105: Adding Dockerfile

naswierczek:
- This will build a container and run all non-integration tests for records-mover


ryantimjohn (changes requested):
@naswierczek this looks great! 

You have a quality job failing: https://app.circleci.com/pipelines/github/bluelabsio/records-mover/3639/workflows/8bb8e207-4583-4766-8991-f9828032bcfd/jobs/48579/parallel-runs/0/steps/0-103

Clean up the additional errors on the MD Linter that you additions added and you should be good to commit!

Additionally, it might be worth us figuring out an auto-formatter instead of a linter so that we don't have to worry about silly errors like these delaying deploying code.

README.md
```diff
@@ -151,3 +151,25 @@ In particular, note:
...
+docker run -it --mount src="$(pwd)/records_mover",target=/records-mover/records_mover,type=bind records-mover
+```
+This will mount your local src overtop of the same directory in the container. Mount any additional directories you
+are working on with additional --mount entries
+
+Alternatively, you can launch the container using docker-compose.
+Alternatively, use `docker-compose up -d` to leverage the docker-compose file to start the container with pre-defined
+mounts.
+```
+docker-comppose up -d
```
ryantimjohn, line 169: Typo: docker-compose

README.md
```diff
@@ -151,3 +151,25 @@ In particular, note:
 * [targets factory methods](https://records-mover.readthedocs.io/en/latest/records_mover.records.targets.html)
 * [move() method](https://records-mover.readthedocs.io/en/latest/records_mover.records.html#records_mover.records.move)
 * [BaseRecordsFormat](https://records-mover.readthedocs.io/en/latest/records_mover.records.html#records_mover.records.base_records_format.BaseRecordsFormat)
+
+## Local Development
+
+The included Dockerfile can be used to build a docker image that is suitable for local development. Mount
+the directory containing your local copy of the repository to have it override what's in the container.
```
ryantimjohn, line 158: Might want to add a line or two about doing docker build. Will need to be named `records-mover:latest` if I'm not wrong about your docker run command below. So

```bash
docker build -t records-mover:latest .
```

or similar

README.md
```diff
@@ -151,3 +151,25 @@ In particular, note:
...
+Alternatively, use `docker-compose up -d` to leverage the docker-compose file to start the container with pre-defined
+mounts.
+```
+docker-comppose up -d
+```
+This will build the container image as necessary, launch it, and mount the most relevant volumes for local dev. 
+The container is hosting a bash shell and will run until you manually shut it down.
+
+Note, if you have to add a dependency to requirements.txt and want to test it locally, add the line
+`COPY requirements.txt .` after `RUN git clone ...` in the Dockerfile and rebuild the container.
```
ryantimjohn, line 175: Should add some lines about local testing here:

Can run unit and type tests in the docker container by running:
```bash
. venv/bin/activate
make citest
make cicoverage
make typecheck
make citypecoverage
```

Then something about the fact that we're figuring out a way to run integration tests locally and that will be added in future builds.

Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM cimg/python:3.9
+ARG VERSION=3.9.1
+ARG NAME=records-mover
+ENV version_env=$VERSION
+ENV name_env=$NAME
+WORKDIR /records-mover
+RUN git clone https://github.com/bluelabsio/records-mover.git .
```
ryantimjohn, line 7: @naswierczek I wonder if it would be better using this:
```dockerfile
COPY . $HOME/records-mover/
```

to use the current directory instead of the existing records-mover dir to build

ryantimjohn: okay, here's a proposed alternative to the cloning from git strategy, this is the whole Dockerfile:

```dockerfile
FROM cimg/python:3.9
ARG VERSION=3.9.1
ARG NAME=records-mover
ENV version_env=$VERSION
ENV name_env=$NAME

USER root

COPY . $HOME/records-mover/
WORKDIR /records-mover

RUN python -m venv venv \
    && . venv/bin/activate \
    && pip install --upgrade pip \
    && pip install --progress-bar=off -r requirements.txt \
    && pip install --upgrade --progress-bar=off 'pandas>=1.5.3,<2' \
    && pip install -e '.[unittest,typecheck]' #
```

naswierczek: @ryantimjohn I considered this at length. Given that this is intended to support local dev, it probably does make sense to assume this dockerfile would be built in a directory where the relevant code was already checked out.

It would make sense to leave it as is if this was being used as part of a build pipeline.
Worth it to change it?

ryantimjohn: one possibility if we want it to be specifically for local dev is renaming the file to `Dockerfile.dev` then add to the docker build statement `docker build -f Dockerfile.dev`. I've seen that in a number of repos where you want specifically a Dockerfile for development and not deployment.

The reason that I brought up building the image from local files rather than what was cloned from the main branch was that I was working on a PR where I had to update the `setup.py` with new build packages, and realized that was very hard with the clone from git hard-coded. Another option might be to have an arg for building it from a particular branch, but having that branch default to main.

Anyway, just some thoughts. This is acceptable as exists aside from the quality issues and maybe adding a little to the readme.

If we do allow building from non-main branches or from the local dir, we should def remove the tests as non-main branches and local files are often not going to pass tests as we're building on them which is the point. =)

naswierczek: RE: Quality issues.
I've fixed all the ones in the README section I added but there are other parts of the README and other MD files getting flagged.

Do you think I fix those as well? They don't seem that severe.

naswierczek:
@ryantimjohn Please review and remove your request for changes when you get a chance! :)

Dockerfile.dev
```diff
@@ -0,0 +1,24 @@
...
+RUN python -m venv venv \
+    && . venv/bin/activate \
+    && pip install --upgrade pip \
+    && pip install --progress-bar=off -r requirements.txt \
+    && pip install --upgrade --progress-bar=off 'pandas>=1.5.3,<2' \
+    && pip install -e '.[unittest,typecheck]' #\
+    && make citest \
+    && make cicoverage \
+    && make typecheck \
+    && make citypecoverage
```
ryantimjohn, line 24: @naswierczek since this is development env, I don't think we should do the checks as the docker build will fail if these fail; sometimes we want to develop on a build that is failing!

ryantimjohn:
Also, you can always hit the re-request review button:

Screenshot 2023-11-14 at 4 38 49 PM

to remove the changes requested status!

Dockerfile.dev
naswierczek: Good call. I moved this to a docker-compose section.

ryantimjohn (approved):
---

ryantimjohn (changes requested):
Ack, sorry, I'm getting an error trying to build this:
```
 > [7/7] RUN python -m venv venv     && . venv/bin/activate     && pip install --upgrade pip     && pip install --progress-bar=off -r requirements.txt     && pip install --upgrade --progress-bar=off 'pandas>=1.5.3,<2'     && pip install -e '.[unittest,typecheck]' #:
#10 0.321 + python -m venv venv
#10 0.407 Error: [Errno 2] No such file or directory: '/records-mover/venv/bin/python'
------
executor failed running [/bin/bash -exo pipefail -c python -m venv venv     && . venv/bin/activate     && pip install --upgrade pip     && pip install --progress-bar=off -r requirements.txt     && pip install --upgrade --progress-bar=off 'pandas>=1.5.3,<2'     && pip install -e '.[unittest,typecheck]' #]: exit code: 1
```

ryantimjohn:
@naswierczek I think I understand what happened, I think you need to add a `.dockerignore` which ignores your local `venv/`, I think this worked for you because you don't have a local venv. I did, which messed up the install step.

ryantimjohn:
This is what my `.dockerignore` looks like (a lot like a .gitignore)
Screenshot 2023-11-20 at 9 42 10 AM

Now builds fine:
Screenshot 2023-11-20 at 9 42 29 AM

README.md
```diff
@@ -151,3 +151,78 @@ In particular, note:
...
+```
+
+The following commands assume the records-mover container has been
+named `records-mover`.
+
+Mount the directory containing your local copy of the repository to
+have it override what's in the container.
+
+```bash
+docker run -it --mount src="$(pwd)/records_mover",target=/records-mover/records_mover,type=bind records-mover
```
ryantimjohn, line 176: I don't know that we need the mounts anymore but tell me if I'm wrong.

ryantimjohn (approved):
Looks great, thanks for addressing all concerns so thoughtfully!

README.md
naswierczek: You need to mount your local files into the container to enable development inside an IDE. Otherwise it is unnecessary.

PR #258, opened 2023-11-13
RM-63 Update for compatibility with Pandas 2

ryantimjohn:


tests/integration/records/multi_db/test_records_table2table.py
```diff
@@ -107,9 +107,14 @@ def move_and_verify(self, source_dbname: str, target_dbname: str) -> None:
...
-                sql = f"DROP TABLE {quoted_target}"
+                drop_target_sql = f"DROP TABLE {quoted_target}"
                 with target_conn.begin():
-                    target_conn.exec_driver_sql(sql)
+                    target_conn.exec_driver_sql(drop_target_sql)
+                quoted_source = quote_schema_and_table(None, source_schema_name, source_table_name,
+                                                       db_engine=source_engine)
+                drop_source_sql = f"DROP TABLE {quoted_source}"
+                with source_conn.begin():
+                    source_conn.exec_driver_sql(drop_source_sql)
```
ryantimjohn, line 117: This is fixing a bug I found while I was in here:

In the table2table job, source tables were not being deleted at the end of the job and therefore thousands of them were stacking up in the long-lived dbs, i.e Redshift and BigQuery.

records_mover/records/pandas/prep_for_csv.py
```diff
@@ -21,13 +21,20 @@ def _convert_series_or_index(series_or_index: T,
                              field: RecordsSchemaField,
                              records_format: DelimitedRecordsFormat,
                              processing_instructions: ProcessingInstructions) -> Optional[T]:
+    if isinstance(series_or_index, pd.Series):
+        try:
+            first_item = series_or_index.iloc[0]
+        except IndexError:
+            first_item = series_or_index.dtype.type()
```
Brunope, line 28: can we not just do this in every case?

tests/integration/records/multi_db/test_records_table2table.py
Brunope: amazing good job

Brunope (approved):

records_mover/records/pandas/prep_for_csv.py
ryantimjohn: Strangely, no. The reason this like 'picking first item' `series_or_index[0]` happens in the first place is that either the data type isn't specified for the series or index _or_ the first item might have a different type from what's specified in the series or index.

ryantimjohn: That's why you have these weird loggers:
```python
            logger.debug("Dtype is %s, first element type %s", series_or_index.dtype,
                         type(first_item))

            logger.warning(f"Found {series_or_index.name} as unexpected "
                           f"type {type(first_item)}")
```

PR #259, opened 2023-12-04
RM-108 remove upper bound on google-api-python-cli

ryantimjohn:


setup.py
```diff
@@ -105,7 +105,7 @@ def initialize_options(self) -> None:
     # 'googleapiclient.discovery.build':
     #
     # https://github.com/googleapis/google-auth-library-python/issues/190
-    'google-api-python-client>=1.8.0,<1.9.0',
+    'google-api-python-client>=1.8.0',
```
Brunope, line 108: maybe a `<2` is in order here for future proofing?

Brunope:
what is the context behind this change?

setup.py
naswierczek: There's a lot of debate about if upper version bounds cause more harm than good. My feeling is that if we don't know that versions >2 will cause problems, we shouldn't add a restriction. Without a comment to add context, we'd quickly lose institutional knowledge about why the bound was there in the first place and it'd become some inscrutable constraint for future us to puzzle over. :)

naswierczek (approved):

ryantimjohn:
@Brunope, to @naswierczek 's point, it's now becoming accepted wisdom in Python package distribution to remove all upper bounds on required packages, then have a system like we do with daily builds + making future warnings trigger failures to catch errors early, then build compatibility with newest package versions wherever possible. Break early and fix. This gives your package the least likelihood of having dependency incompatibilities with other packages that people need in their workflows. Only where there's a genuine inability to make your package incompatible with most recent versions should you have upper bounds.

I'm going through all of our upper bounds now one-by-one and seeing if we can reasonably remove them.

Here's an article about this:
https://iscinumpy.dev/post/bound-version-constraints/

ryantimjohn:
Screenshot 2023-12-11 at 11 12 56 AM
@naswierczek why'd you remove you supportive comment =p

naswierczek:
@ryantimjohn I certainly didn't do it on purpose!

PR #265, opened 2023-12-11
RM-113 remove upper bound pyyaml

ryantimjohn:


setup.py
```diff
@@ -298,15 +298,7 @@ def initialize_options(self) -> None:
...
-          # awscli pins PyYAML. We have seen issues at
-          # run-time if we don't constrain things here as well. So
-          # this tracks the awscli req for python >= 3.6 which is
-          # what we support
-          #
-          # https://github.com/aws/aws-cli/blob/develop/setup.py
-          # #### NEW UNFORTUNATE CONSTRAINTS #####
-          # https://github.com/yaml/pyyaml/issues/724
-          'PyYAML>=3.10,<=5.3.1',
+          'PyYAML>=3.10',
```
Brunope, line 301: pyyaml was one of the packages i remember that was causing problems on some specific versions when I was recently fixing homebrew stuff. which version did this test run with?

ryantimjohn: This ran across 3.8, 3.9, 3.10, and 3.11

ryantimjohn: Do you have any logs from the home brew failure? Also, sorry, I didn't purposefully merge this but seems like I merged the PR after it so this merged 🙃

Brunope: no which version of pyyaml i mean. but anyway what i ended up having to do was increase the pyyaml version to fix the brew build so i think this is fine

PR #266, opened 2023-12-11
RM-114 remove upper bounds on db-facts

ryantimjohn:


Brunope:
damn looks like there is indeed some issue with the 1password db-facts  change which is unsettling. the test is simply timing out

Brunope:
nvm it was the gcloud thing, excellent

ryantimjohn:
Also, I ran too many simultaneous checks yesterday and they overwhelmed our very low resource redshift itest instance @Brunope

Brunope:
ah okay that makes sense why it was seeing redshift timeout errors

Brunope (approved):

PR #269, opened 2023-12-13
RM-118: Removing deprecrated dependency

naswierczek:
- libraries that used to rely on it have already been updated

[Jira ticket](https://bluelabsio.atlassian.net/browse/RM-118)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -180,7 +180,6 @@ aws_dependencies = [
 gsheet_dependencies = [
     'google',
     'google_auth_httplib2',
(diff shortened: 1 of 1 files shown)


naswierczek:
@ryantimjohn @Brunope Thoughts on just removing this dependency entirely? The library is deprecated and has been replaced. I checked our code and I can't see where we reference it directly.

Brunope (approved):

ryantimjohn (approved):

PR #273, opened 2023-12-20
remove system_packages, readthedocs no longer supports it

Brunope:


naswierczek (approved):

Brunope:
previously this flag told the readthedocs builders to use existing pre-installations of packages like numpy to speed up the build process. They removed support for this flag earlier this year, and require all users to install things via requirements.txt. All our dependencies to build docs should already be in setup.py with the "docs" extra so this should continue to work without the flag, just slower.

PR #282, opened 2024-02-08
RM-124: Add command to perform a healthcheck against airbyte server

naswierczek:
- adds new optional arugment that can be passed to mvrec that will query the healthcheck endpoint on the configured airbyte instance and return either true or false depending on if that endpoint returns a 200 status code
- Added new class to support argument
- Added unit tests for new class method


records_mover/records/airbyte/airbyte.py
```diff
@@ -0,0 +1,45 @@
...
+                     If a session isn't provided, one will be requested
+        """
+        if session is None:
+            self.session = Session()
+        else:
+            self.session = session
+
+    def healthcheck(self) -> bool:
+        self.session.set_stream_logging()
+        data = self.session.creds.airbyte("local-airbyte")
```
naswierczek, line 29: This can't be hard-coded to local-airbyte

Brunope: what should it be? required argument on the CLI? i'm okay with it defaulting to this maybe

records_mover/records/cli.py
```diff
@@ -114,7 +116,17 @@ def main() -> None:
...
+    if args.healthcheck:
+        try:
+            engine = AirbyteEngine()
+            result = engine.healthcheck()
+            if result:
+                print("Airbyte Status: OK!")
+            else:
+                print("Airbyte Status: Unhealthy")
+        except Exception:
+            sys.exit(1)
```
Brunope, line 128: can we instead let the exception propagate so it's visible to the user? it'll sys exit anyway

records_mover/records/airbyte/airbyte.py
naswierczek: Required argument doesn't make a ton of sense so having just a single stable entry we define is enticing. In production, we'll probably only have a single stable hostname for the airbyte instance. We'll have a different instance to run integration tests against and that's about it.

naswierczek:
@Brunope This is just step one of an overall larger project to integrate airbyte as the engine for doing stuff. Anything else you'd want to see in this PR before it merges?  I'm feeling like I should put in a basic feature flag where if you don't have an environment variable set, the new option doesn't even appear.

Brunope (approved):
sorry i meant to click approve like last week but i guess forgot - i think this is going in the right general direction, good to merge once all the checks are passing

PR #287, opened 2024-03-14
RM-126: Fixing healthcheck lookup

naswierczek:
- Swap out how healthcheck argument is retrieved to use a safer method


Brunope (approved):
can't you just do sometihng like `if args.healthcheck`? isnt that the whole point of argparse?

naswierczek:
Typically I could but I decided at the last minute to throw the airbyte functionality behind a feature flag so that we can safely release records mover whenever we need to without needing to tell anyone. The consequence of how I implemented that is args.healthcheck may not always exist which would throw an exception I'd have to catch anyway. :)

PR #293, opened 2024-05-01
RM-130 test clean build

Brunope:


pyproject.toml
```diff
@@ -1,3 +1,5 @@
+# temp comment to bump build
```
Brunope, line 1: woops ill remove this before i merge, i dont wavnna trigger a whole new build though

.circleci/config.yml
```diff
@@ -72,10 +70,11 @@ commands:
                 pip install --force-reinstall -v --progress-bar=off 'sqlalchemy<<parameters.sqlalchemy_version>>'
               fi
             fi
-      - save_cache:
-          key: deps-v11-<<parameters.python_version>>-<<parameters.pandas_version>>-<<parameters.extras>>-<<parameters.include_dev_dependencies>>-{{ .Branch }}-{{ checksum "requirements.txt" }}-{{ checksum "setup.py" }}-{{ checksum "pyproject.toml" }}
-          paths:
-            - "venv"
+            echo "***********************************************"
+            echo "REQUIREMENTS **********************************"
```
naswierczek, line 74: Is something missing here or is this just so there's something very obvious to find to mark a specific spot when reviewing the logs?

tests/integration/records/single_db/numeric_expectations.py
```diff
@@ -180,8 +180,8 @@
         'uint64': 'NUMERIC(20, 0)',
         'float16': 'REAL',
         'float32': 'REAL',
-        'float64': 'DOUBLE PRECISION',
-        'float128': 'DOUBLE PRECISION',  # Postgres doesn't support >float64
+        'float64': 'DOUBLE_PRECISION',
```
naswierczek, line 183: What's the purpose of this change?

I can't say it's right or wrong but it's the only change that doesn't "fit" with the others in this PR.

naswierczek (approved):
I left a question on the DOUBLE PRECISION -> DOUBLE_PRECISION change for my own edification.

Please make sure to remove that comment from pyproject.toml before merge! :)

LGTM

tests/integration/records/single_db/numeric_expectations.py
Brunope: Yeah, this is a result of downgrading sqlalchemy to 1.4, where i guess it returns a sligthly different string here

PR #296, opened 2024-06-17
RM-133: cap numpy < 2.0

Brunope:


Brunope:
not super happy with how the pandas_version and numpy_version are specified in these tests. i feel like i should only have had to update the pyproject.toml for this change. But a bunch of tests are explicitly installing pandas which then requires explicitly installing numpy < 2 when i dont *think* pandas should be required at all?

naswierczek (approved):

naswierczek:
pandas is installed as a dependency for unittest. [pyproject.toml](https://github.com/bluelabsio/records-mover/blob/main/pyproject.toml#L186) Further complicating the issue is that pandas can totally be installed, and parts of it will function just fine, without numpy installed.

I'm not going to dig further to work out why/how pandas is being used in the tests but that's why I suspect it's installed.

